PR #4, opened 2019-01-03
fix model pub var name bug, and add UI

ztimc:


ztimc:
Next Commit, I want separate Tinodesdk, How do you feel

ios/MessageViewController.swift
```diff
@@ -0,0 +1,20 @@
...
+//  MessageViewController.swift
+//  ios
+//
+//  Created by ztimc on 2019/1/3.
+//  Copyright © 2019 Tinode. All rights reserved.
+//
+
+import UIKit
+
+class MessageViewController: UIViewController {
```
aforge: Regarding this view: in Tindroid the UI for it is implemented manually.
I am not convinced the same should be done here but I'm open to discussion.
One thing I've looked into is MessageKit. It seems to mostly satisfy our needs here. The tradeoff of course is that we will become tied to MessageKit and its codebase.
Do you have an opinion on this?

PR #8, opened 2019-04-12
Online/offline updates in real time

or-else:
I've added processing of {pres} to 'me' and updated Tinode class to be a bit closer to the java version.


or-else:
@aforge pls take a look

Tinodios/Supporting Files/Base.lproj/LaunchScreen.storyboard
```diff
@@ -11,15 +15,27 @@
...
                 <viewController id="01J-lp-oVM" sceneMemberID="viewController">
                     <view key="view" contentMode="scaleToFill" id="Ze5-6b-2t3">
-                        <rect key="frame" x="0.0" y="0.0" width="375" height="667"/>
+                        <rect key="frame" x="0.0" y="0.0" width="414" height="896"/>
                         <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
-                        <color key="backgroundColor" red="1" green="1" blue="1" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
+                        <subviews>
+                            <imageView clipsSubviews="YES" userInteractionEnabled="NO" contentMode="scaleAspectFit" horizontalHuggingPriority="251" verticalHuggingPriority="251" fixedFrame="YES" image="ios-logo-1024" translatesAutoresizingMaskIntoConstraints="NO" id="fnw-Nz-LWI">
+                                <rect key="frame" x="87" y="328" width="240" height="240"/>
+                                <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMaxY="YES"/>
```
aforge: We'll need to remove top and left autoresizing constraints.
Otherwise, the logo slips down to the bottom of the screen.

TinodeSDK/Topic.swift
```diff
@@ -432,9 +455,10 @@ open class Topic<DP: Codable, DR: Codable, SP: Codable, SR: Codable>: TopicProto
         let name = self.name
         var newTopic = false
         if tinode!.getTopic(topicName: name) == nil {
-            tinode!.registerTopic(topic: self)
```
aforge: What's the reason we don't want to register the topic here?

TinodeSDK/Topic.swift
```diff
@@ -467,7 +491,8 @@ open class Topic<DP: Codable, DR: Codable, SP: Codable, SR: Codable>: TopicProto
                 if let e = err as? TinodeError, newTopic {
                     if case TinodeError.serverResponseError(let code, _, _) = e {
                         if code >= 400 && code < 500 {
-                            self?.tinode?.unregisterTopic(topicName: name)
+                            self?.tinode?.stopTrackingTopic(topicName: name)
+                            self?.persist(false)
```
aforge: It looks like every call to stopTrackingTopic is accompanied by persist(false)
What's the reason for decoupling?

aforge:
@or-else I pushed a couple changes to devel. Please merge into contacts branch.

aforge:
Also, how do we want to present the UI in the landscape mode? (you know you can test it right in XCode without even running the app on a simulator, right?)
We either need to restrict it to portrait-only or change the layout to adapt.

Tinodios/ChatListViewController.swift
```diff
@@ -55,10 +58,23 @@ class ChatListViewController: UITableViewController, ChatListDisplayLogic {
 
     func displayChats(_ topics: [DefaultComTopic]) {
         self.topics = topics
+        self.rowIndex.removeAll()
+        var i = 0
+        for topic in topics {
```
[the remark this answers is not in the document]
or-else: If I do that it complains: `Referencing initializer 'init(uniqueKeysWithValues:)' on 'Dictionary' requires that 'DefaultComTopic' (aka 'ComTopic<VCard>') conform to 'Hashable'`

I simplified it a bit though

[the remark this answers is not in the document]
or-else: OK, figured it out. Swift is different :)

Tinodios/ChatListViewController.swift
```diff
@@ -70,9 +70,11 @@ class ChatListViewController: UITableViewController, ChatListDisplayLogic {
     }
 
     func updateChat(_ name: String) {
+        print("updateChat "+name)
```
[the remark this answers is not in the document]
or-else: Removed - was used for debugging. Will keep in in mind.

Tinodios/ChatListViewController.swift
```diff
@@ -70,9 +70,11 @@ class ChatListViewController: UITableViewController, ChatListDisplayLogic {
...
         guard let position = rowIndex[name] else { return }
+        print("updateChat at "+String(position))
```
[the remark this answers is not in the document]
or-else: Also removed

Tinodios/Supporting Files/Base.lproj/LaunchScreen.storyboard
or-else: I don't see it when I run it. Anyway, I added some constraints (aspect ratio, centered vertically/horizontally, margins). It should be fine now.

TinodeSDK/Topic.swift
or-else: These are the changes in sync with Android: topic registration was moved to topic initializer to be sure all created topics are always registered.

TinodeSDK/Topic.swift
or-else: I hit this problem on Android and expect it to appear here as well: when the app received a push notification, it needs to instantiate a topic in order to get topic's display name and avatar. It does not need to be in memory once the push is processed, but it should not be removed from storage.

or-else:
> @or-else I pushed a couple changes to devel. Please merge into contacts branch.

Done

or-else:
> Also, how do we want to present the UI in the landscape mode? (you know you can test it right in XCode without even running the app on a simulator, right?)
> We either need to restrict it to portrait-only or change the layout to adapt.

OK, will make sure they are OK in landscape mode and on smaller screens.

aforge:
Thanks. This is good. Let's merge this PR into devel.

PR #12, opened 2019-04-22
Messages layout, DB changes, refactored toast

or-else:
* Messages VC refactored to show messages in a nicer way (still needs work)
* SubscriptionDb is fully functional now
* .table is now non-optional
* Database is versioned, will drop/create tables and indexes on version changes.
* Toast is made more iOS-like


Tinodios/MessageViewController.swift
```diff
@@ -157,39 +185,61 @@ extension MessageViewController: MessagesDataSource {
...
             NSAttributedString.Key.foregroundColor: UIColor.gray
             ])
     }
 }
 
 extension MessageViewController: MessagesDisplayDelegate, MessagesLayoutDelegate {
+    // Helper checks.
+
+    func isPreviousMessageSameSender(at indexPath: IndexPath) -> Bool {
+        guard indexPath.section - 1 >= 0 else { return false }
```
aforge: indexPath.section > 0 is easier to read.

Tinodios/MessageViewController.swift
```diff
@@ -54,32 +56,56 @@ class MessageViewController: MessageKit.MessagesViewController, MessageDisplayLo
...
+
         messagesCollectionView.addSubview(refreshControl)
         refreshControl.addTarget(self, action: #selector(loadNextPage), for: .valueChanged)
-    }
-
-    override func viewDidLoad() {
-        super.viewDidLoad()
 
         if let layout = messagesCollectionView.collectionViewLayout as? MessagesCollectionViewFlowLayout {
+            layout.sectionInset = UIEdgeInsets(top: 1, left: 8, bottom: 1, right: 8)
```
aforge: Consider moving it to a separate function (e.g. configureLayout).

Tinodios/MessageViewController.swift
```diff
@@ -157,39 +185,61 @@ extension MessageViewController: MessagesDataSource {
...
     func configureAvatarView(_ avatarView: AvatarView, for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) {
         // Hide current user's avatar as well as peer's avatar in p2p topics.
         // Avatars are useful in group topics only
-        if topic!.isGrpType || isFromCurrentSender(message: message) {
-            avatarView.isHidden = true
-            return
-        }
-
+        avatarView.isHidden = topic!.isP2PType || isNextMessageSameSender(at: indexPath)
         let sub = topic?.getSubscription(for: message.sender.id)
```
aforge: if let sub = topic?.getSubscription(...) {
   ... // This way you unwrap the sub optional once (and don't need to do it again).
}

Tinodios/MessagesFlowLayout.swift
```diff
@@ -0,0 +1,67 @@
...
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+ */
+
+import Foundation
+import MessageKit
+
+open class MessagesFlowLayout: MessagesCollectionViewFlowLayout {
```
aforge: Why open? Are you planning on making it available to another package?

Tinodios/MessagesFlowLayout.swift
```diff
@@ -0,0 +1,67 @@
+/*
```
aforge: Why do we need license here?

Tinodios/MessageViewController.swift
or-else: done

Tinodios/MessageViewController.swift
or-else: done

Tinodios/MessageViewController.swift
or-else: done

Tinodios/MessagesFlowLayout.swift
or-else: open removed, the calculator class is made fileprivate

Tinodios/MessagesFlowLayout.swift
or-else: done

Tinodios/UiUtils.swift
```diff
@@ -67,33 +67,54 @@ class UiUtils {
...
+        icon.tintColor = UIColor.white
+        icon.frame = CGRect(x: spacing, y: spacing, width: iconSize, height: iconSize)
+
+        label.frame = CGRect(
+            x: iconSize + spacing * 2, y: (messageHeight - label.frame.height) / 2,
+            width: label.frame.width + spacing * 2, height: label.frame.height)
+
+        let toastView = UIView()
+        toastView.alpha = 0
+        toastView.backgroundColor = UIColor.red.withAlphaComponent(0.6)
```
aforge: Do we always want to display it in red?
I think certain information/warning messages may be better with a more neutral background.
So, do you want to make  the background color configurable (e.g. function param)? Or pass in message severity (INFO, WARNING, ERROR)?

Tinodios/db/SubscriberDb.swift
```diff
@@ -182,49 +182,53 @@ class SubscriberDb {
...
+        guard let udb = BaseDb.getInstance().userDb else { return nil }
+        guard let tdb = BaseDb.getInstance().topicDb else { return nil }
+        let joinedTable = self.table.select(
+            self.table[self.id],
             self.topicId,
             self.userId,
-            self.status,
-            self.mode,
-            self.updated,
+            self.table[self.status],
```
aforge: Why do you need this change? self.status refers to the status column.

Tinodios/db/BaseDb.swift
```diff
@@ -9,6 +9,10 @@ import Foundation
 import SQLite
 
 public class BaseDb {
+    // Current database schema version. Increment on schema changes.
+    public static let kSchemaVersion = 100
```
aforge: public static let kSchemaVersion: Int32 = 100
This way you don't need to typecast it in the code below.

Tinodios/db/BaseDb.swift
```diff
@@ -57,23 +61,30 @@ public class BaseDb {
...
-        self.accountDb!.createTable()
         self.userDb = UserDb(self.db!)
-        self.userDb!.createTable()
         self.topicDb = TopicDb(self.db!)
-        self.topicDb!.createTable()
         self.subscriberDb = SubscriberDb(self.db!)
-        self.subscriberDb!.createTable()
         self.messageDb = MessageDb(self.db!)
+
+        if self.db?.schemaVersion != Int32(BaseDb.kSchemaVersion) {
```
aforge: Can you change it to self.db!.schemaVersion please?

TinodeSDK/Topic.swift
```diff
@@ -510,8 +510,10 @@ open class Topic<DP: Codable, DR: Codable, SP: Codable, SR: Codable>: TopicProto
     }
 
     private func loadSubs() -> Int {
-        // todo: implement
-        return 0
+        guard let loaded = store?.getSubscriptions(topic: self) else { return 0 }
+        subs = (Dictionary(uniqueKeysWithValues: loaded.map { ($0.user, $0) }) as! [String : Subscription<SP, SR>])
```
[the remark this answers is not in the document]
or-else: Yes, it's also needed. But it's unused so far, so I'll add it later together with the code which will use it.

Tinodios/UiUtils.swift
or-else: I agree, but I would rather add it when it's needed. I generally prefer not to generalize too early.

Tinodios/db/SubscriberDb.swift
or-else: Because just `status` is ambiguous. There is `topics.status` too and join barfs at having unattributed `status`.

Tinodios/db/BaseDb.swift
or-else: done

Tinodios/db/BaseDb.swift
or-else: done

Tinodios/db/SubscriberDb.swift
or-else: `self.status` translates into sql `"status"`
` self.table[self.status]` -> `"subscriptions"."status"`

Tinodios/UiUtils.swift
or-else: And it's not just the colors, it's also the icons like (!) /!\ (i).

TinodeSDK/Topic.swift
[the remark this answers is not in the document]
or-else: Actually, no harm in doing it now. Added.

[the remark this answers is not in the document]
or-else: Interesting. There is the same bug on Android: subsUpdated is never assigned or used for 'me' topic.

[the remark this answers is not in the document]
or-else: And it's not correctly used for all other topics as well, both Android & iOS :)

Tinodios/UiUtils.swift
aforge: Sounds good to me.

Tinodios/db/SubscriberDb.swift
aforge: Makes sense. Thanks for clarifying.

Tinodios/db/BaseDb.swift
```diff
@@ -57,23 +61,30 @@ public class BaseDb {
...
         self.subscriberDb = SubscriberDb(self.db!)
-        self.subscriberDb!.createTable()
         self.messageDb = MessageDb(self.db!)
+
+        if self.db!.schemaVersion != BaseDb.kSchemaVersion {
+            print("Schema has changed from \(self.db?.schemaVersion ?? -1) to \(BaseDb.kSchemaVersion)")
+            // Delete database if schema has changed.
+            self.dropDb()
+
+            self.db?.schemaVersion = BaseDb.kSchemaVersion
```
aforge: Force unwrap self.db here as well please.

or-else: done

PR #13, opened 2019-04-23
Message formatting

or-else:
Looks much nicer now: 
Simulator Screen Shot - iPhone Xʀ - 2019-04-23 at 19 19 35


TinodeSDK/Tinode.swift
```diff
@@ -285,13 +292,23 @@ public class Tinode {
...
             }
         } else if let meta = serverMsg.meta {
+            let updated: Date?
             if let t = getTopic(topicName: meta.topic!) {
                 //t.route
                 t.routeMeta(meta: meta)
+                updated = t.updated
             } else {
-                _ = maybeCreateTopic(meta: meta)
+                let t = maybeCreateTopic(meta: meta)
```
aforge: if let t = maybeCreateTopic(meta: meta) {
    self.updated = t.updated
}

Tinodios/MessageViewController.swift
```diff
@@ -172,6 +160,8 @@ extension MessageViewController {
     func displayChatMessages(messages: [StoredMessage]) {
         self.messages = messages.reversed()
         self.messagesCollectionView.reloadData()
+        // FIXME: don't scroll to bottom in response to loading the next page.
+        // Maybe don't scroll to bottom if the view is not at the bottom already.
```
aforge: Right. We need to know the current state of the VC.
Scrolling to bottom is only needed upon the first load of the data.

README.md
```diff
@@ -1,45 +1,36 @@
-# Tinodios
+# Tinodios: Tinode Messaging Client for iOS 
 
-Experimental iOS client.
+iOS client for [Tinode](https://github.com/tinode/chat) in Swift.
```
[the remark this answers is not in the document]
or-else: :)

TinodeSDK/Tinode.swift
or-else: done

PR #16, opened 2019-04-24
delivery indicators work now

or-else:


or-else:
This is weird. Here is what I see:
Simulator Screen Shot - iPhone Xʀ - 2019-04-25 at 17 36 00

The tint is supposed to be either grey or blue. Let me check why it may be different in your simulator.

aforge:
Strange. It's very unpredictable for me.
Sometimes, it shows all the indicators as intended. But after I post a new message, some of these indicators disappear (seemingly in a random fashion).

or-else:
Fixed. 

The problem was because I missed the fact that `UICollectionView` recycles the views, i.e. if an instance of `UIImageView` moves off the screen then `UICollectionView` may reuse that instance to show another avatar for a different message. If that `UIImageView` was previously `.isHidden=true`, then the new avatar would not show. The fix is to set `.isHidden` explicitly every time instead of assuming it's `false`.

Tinodios/MessageViewController.swift
```diff
@@ -237,23 +240,46 @@ extension MessageViewController: MessagesDisplayDelegate, MessagesLayoutDelegate
...
+        if isFromCurrentSender(message: message) {
+            // Use current user's avatar for showing delivery markers.
+            guard let msg = message as? StoredMessage, let topic = topic else { return }
+
+            let iconName: String
+            var tint: UIColor = MessageViewController.kDeliveryMarkerColor
+            if msg.status == nil || msg.status! <= BaseDb.kStatusSending {
+                iconName = "outline_schedule_white_48pt"
+            } else {
+                if topic.msgReadCount(seq: msg.seq) > 0 {
```
aforge: We should perhaps cache the results here (by making this info a part of StoredMessage).

aforge:
I understand this PR is for display changes only, correct?

Tinodios/MessageViewController.swift
or-else: Then it will require cache invalidation. The operation is not that expensive: it runs for displayed messages only.
Maybe it would be useful to have another method `hasReadMessages` which will return on finding the first subscriber who red the message instead of iterating over all of them.

or-else:
> I understand this PR is for display changes only, correct?

If you mean the delivery marker is not updated in real time then yes, it's not done yet.

PR #17, opened 2019-04-26
Drafty parsing works

or-else:
This is the first part of implementing Drafty: parsing works, messages are converted to Drafty and sent as such. They appear formatted in Tindroid and Webapp.

Tinodios shows Drafty messages as plain text.


TinodeSDK/model/Drafty.swift
```diff
@@ -7,23 +7,874 @@
...
+    public static func parse(content: String) -> Drafty {
+        // Break input into individual lines because format cannot span multiple lines.
+        // This breaks lines by \n only, we do not expect to see Windows-style \r\n.
+        let lines = content.components(separatedBy: .newlines)
+        // This method also accounts for Windows-style line breaks, but it's probably not needed.
+        // let lines = content.split { $0 == "\n" || $0 == "\r\n" }.map(String.init)
+        var blks: [Block] = []
+        var refs: [Entity] = []
+
+        var spans: [Span]?
```
aforge: You can declare it within the loop so you don't need to make it optional.

aforge:
I'm also getting an error when processing an incoming message:

in: {"data":{"topic":"usrC4XXLB5_HJ0","from":"usrC4XXLB5_HJ0","ts":"2019-04-24T00:39:20.511Z","seq":1,"content":"Crazee Edeee, his prices are INSANE!!!"}}
onMessage error: typeMismatch(Swift.Dictionary<Swift.String, Any>, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "data", intValue: nil), CodingKeys(stringValue: "content", intValue: nil)], debugDescription: "Expected to decode Dictionary<String, Any> but found a string/data instead.", underlyingError: nil))

You'll need to add a decoding constructor to Drafty.

TinodeSDK/model/ClientMessages.swift
```diff
@@ -347,13 +347,14 @@ public class MsgClientNote: Encodable {
...
 
-public class MsgClientPub: Encodable {
+public class MsgClientPub<C: Encodable>: Encodable {
     let id: String?
     let topic: String?
     let noecho: Bool?
     let head: [String:JSONValue]?
-    let content: JSONValue?
-    init(id: String?, topic: String?, noecho: Bool?, head: [String:JSONValue]?, content: JSONValue?) {
+    let content: C
```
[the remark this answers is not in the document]
or-else: It could be either String or Drafty as your comment

> I'm also getting an error when processing an incoming message

points out. But yes, what you are suggesting is a good idea. I'll write a custom decoder which can generate Drafty from either string or json and get rid of `<C>`

TinodeSDK/model/Drafty.swift
or-else: It's not that simple. Both `toTree` and `chunkify` return `[Span]?` which is assigned to the same variable `spans` in the loop. I'll see if it's worth refactoring them to get rid of optionals in the returns.

or-else:
> This is a perfect situation where unit tests might come in handy.
> What was your testing strategy with Android and Web?

There is none. I'll check it out.

or-else:
> I'm also getting an error when processing an incoming message:
> 
> You'll need to add a decoding constructor to Drafty.

Yep.

TinodeSDK/model/Drafty.swift
or-else: don

TinodeSDK/model/Drafty.swift
```diff
@@ -7,23 +7,874 @@
...
+        var entityMap: [String:JSONValue] = [:]
+        for line in lines {
+            spans = []
+            // Select styled spans.
+            for i in 0..<Drafty.kInlineStyleName.count {
```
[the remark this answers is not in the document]
or-else: Done. Differently but quite swiftly.

TinodeSDK/model/JSONValue.swift
```diff
@@ -49,8 +52,19 @@ public enum JSONValue: Codable {
...
             throw DecodingError.typeMismatch(JSONValue.self, DecodingError.Context(codingPath: container.codingPath, debugDescription: "Not a JSON"))
         }
     }
+
+    public func asInt() -> Int {
+        switch self {
+        case .int(let val):
+            return val
+        default:
+            return nil! // Crash and burn.
```
aforge: assert() with an error message is a better choice here.

Tinodios/MessageViewController.swift
```diff
@@ -145,7 +145,7 @@ extension StoredMessage: MessageType {
         get { return self.ts ?? Date() }
     }
     var kind: MessageKind {
-        get { return .text(self.content ?? "") }
+        get { return .text(self.content?.string ?? "") }
```
aforge: Optional: maybe I'm rushing here a bit but you could
{
  guard let content = self.content else { return .text("") }
  return content.isPlain ? .text(content.string) : .attributedString(content.string)
}

TinodeSDK/model/JSONValue.swift
or-else: I actually removed it, just did not merge the change. I'll update it.

aforge:
> > This is a perfect situation where unit tests might come in handy.
> > What was your testing strategy with Android and Web?
> 
> There is none. I'll check it out.

It'd be good. I played a bit with Android and Web and it was unclear to me what certain inputs should have produced: e.g. aaa *bb _123_ b* aaa
I'd expect "aaa <b>bb</b> <i>123</i> <b>b</b> aaa"
But what I was getting is "aaa <b>bb</b> <i>123</i> aaa" (second "b" disappeared).

Tinodios/MessageViewController.swift
or-else: I added the guard statement but `.attributedText()` is premature. The `Drafty.toAttributed()` does not work yet. I'm writing it in the `drafty` branch.

or-else:
> `aa *bb _123_ b* aaa`

Fixed.

As far as I can tell, Drafty is clean.

PR #23, opened 2019-05-23
Swipe left in ChatListView to delete topic

or-else:


TinodeSDK/Topic.swift
```diff
@@ -771,6 +771,21 @@ open class Topic<DP: Codable, DR: Codable, SP: Codable, SR: Codable>: TopicProto
         listener?.onMeta(meta: meta)
     }
 
+    /// Delete topic
+    @discardableResult
+    public func delete() -> PromisedReply<ServerMessage>? {
+        // Delete works even if the topic is not attached.
+        return try! tinode!.delTopic(topicName: name)?.then(
+            onSuccess: { [weak self] msg in
```
aforge: No need for [weak self] here.

or-else: done

PR #29, opened 2019-05-28
Messages

or-else:
Refactoring Find view and New Chats view to use common table cells + some other UI tweaks.


Tinodios/FindViewController.swift
```diff
@@ -181,3 +171,17 @@ extension FindViewController: UISearchResultsUpdating, UISearchControllerDelegat
...
+    func selected(from cell: UITableViewCell) {
+        guard let indexPath = tableView.indexPathForSelectedRow else { return }
+        guard let id = getUniqueId(for: indexPath) else { return }
+        // If the search bar is active, deactivate it.
+        if searchController.isActive {
+            DispatchQueue.main.async {
+                self.searchController.isActive = false
+            }
+        }
+        presentChatReplacingCurrentVC(with: id)
```
aforge: The problem is it doesn't work when you type something in the search bar:
self.searchController.isActive = false makes the view reload the list of contacts and the presentChatReplacingCurrentVC() call fails for some reason.
Maybe try to not trigger a reload when cancelling the search bar in this specific case?

Tinodios/NewGroupViewController.swift
```diff
@@ -98,13 +109,31 @@ class NewGroupViewController: UIViewController, UITableViewDataSource {
         let tagsList = Utils.parseTags(from: tags)
         self.interactor?.createGroupTopic(titled: groupName, subtitled: privateInfo, with: tagsList, consistingOf: members, withAvatar: avatar)
     }
+
+    /// Show message that no members are selected.
+    public func toggleNoSelectedMembersNote(on show: Bool) {
```
aforge: Can it be private?

Tinodios/NewGroupViewController.swift
```diff
@@ -98,13 +109,31 @@ class NewGroupViewController: UIViewController, UITableViewDataSource {
...
+        if show {
+            let messageLabel = UILabel(frame: CGRect(x: 0, y: 0, width: selectedCollectionView.bounds.size.width, height: selectedCollectionView.bounds.size.height))
+            messageLabel.text = "No members selected"
+            messageLabel.textColor = .gray
+            messageLabel.numberOfLines = 0;
```
aforge: No need for semicolons (it's Swift) ;-)

Tinodios/widgets/RoundImageView.swift
```diff
@@ -75,21 +100,36 @@ class RoundImageView: UIImageView {
...
         } else {
-            self.letterTileFont = UIFont.preferredFont(forTextStyle: .title2)
-            let (fg, bg) = RoundImageView.selectBackground(id: id, dark: Tinode.topicTypeByName(name: id) == TopicType.p2p)
-            self.letterTileTextColor = fg
-            self.backgroundColor = bg
-
-            // Avatar placeholder.
-            var firstChar = title ?? ""
-            firstChar = firstChar.isEmpty ? "?" : firstChar
+            if id != nil && !id!.isEmpty {
```
aforge: A more swifty way would be
if let id = id, !id.isEmpty {}

or-else: done

Tinodios/NewGroupViewController.swift
or-else: done

Tinodios/NewGroupViewController.swift
or-else: done

Tinodios/FindViewController.swift
or-else: fix

PR #32, opened 2019-07-12
Messages

or-else:
This contains the fix for the bug in message display and some initial work for image previews.


Tinodios/Utils.swift
```diff
@@ -36,32 +36,40 @@ class Utils {
...
         var removed: [Int] = []
+        var mutated: [Int] = []
 
         // Match old array against the new array to separate removed items from inserted.
         var o = 0, n = 0
         while o < old.count || n < new.count {
             if o == old.count || (n < new.count && old[o].seqId > new[n].seqId) {
                 // Present in new, missing in old: added
                 inserted.append(n)
+                if mutated.count == 0 || mutated[mutated.count - 1] != n {
```
aforge: if (mutated.last ?? -1) != n

or-else: Nice! I did not know about `.last`

PR #38, opened 2019-07-18
Fixes for Archived Chats link, silhouette icon incorrectly used in new group topics

or-else:
* Archived Chats link converted from a table cell to a footer.
* Placeholder silhouette icon was incorrectly used as a new group topic avatar & new user avatar when no avatar image was uploaded.
* In group topic member list current user's avatar would show (Y) (from You) instead of the first letter of the name.
* Group topics were not added to subscriptions cache in FndTopic

### DIFF
--- a/Tinodios/ChatListViewController.swift
+++ b/Tinodios/ChatListViewController.swift
@@ -17,6 +17,8 @@ protocol ChatListDisplayLogic: class {
 
 class ChatListViewController: UITableViewController, ChatListDisplayLogic {
 
+    private static let kFooterHeight: CGFloat = 30
+
     @IBOutlet var chatListTableView: UITableView!
 
     var interactor: ChatListBusinessLogic?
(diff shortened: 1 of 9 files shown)


Tinodios/ChatListViewController.swift
```diff
@@ -174,15 +189,10 @@ extension ChatListViewController {
...
     override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
         tableView.deselectRow(at: indexPath, animated:  true)
-        let section = indexPath.section
-        if self.numArchivedTopics > 0 && section == 0 {
-            self.performSegue(withIdentifier: "Chats2Archive", sender: nil)
-        } else {
-            self.performSegue(withIdentifier: "Chats2Messages",
-                              sender: self.topics[indexPath.row].name)
-        }
+        self.performSegue(withIdentifier: "Chats2Messages", sender: self.topics[indexPath.row].name)
```
aforge: I guess you can remove this and drag a segue directly from the table cell proto to MessageVC in the storyboard.
The reason I added it was being able to segue to different VCs depending on the section.

or-else: I don't see how to do it. The cell is in a separate xib. I don't see an option to add a segue to xib.

aforge: You are right. The cell has moved to a separate xib, indeed. So let's keep as it is.

PR #40, opened 2019-07-19
Bug fixes

or-else:
* Fix for a crash in ChatListViewController
* Removed FindRouter because it's unused
* Fix for (sometimes) failing transitions from FindView to MessageView. Still not great because it takes ~200-300 ms to transition.
* Assigned MessageVC to rootController on basic login too


Tinodios/FindInteractor.swift
```diff
@@ -30,7 +30,7 @@ class FindInteractor: FindBusinessLogic {
 
     static let kTinodeImProtocol = "Tinode"
     var presenter: FindPresentationLogic?
-    var router: FindRoutingLogic?
+    // var router: FindRoutingLogic?
```
aforge: Remove?

Tinodios/LoginViewController.swift
```diff
@@ -59,7 +59,12 @@ class LoginViewController: UIViewController {
             let storyboard = UIStoryboard(name: "Main", bundle: nil)
             let destinationVC = storyboard.instantiateViewController(withIdentifier: "ChatsNavigator") as! UINavigationController
 
-            self.show(destinationVC, sender: nil)
+            if let window = UIApplication.shared.keyWindow {
+                // Use ChatList as the new root view controller (used in navigation).
+                window.rootViewController = destinationVC
+             }
+
+            // self.show(destinationVC, sender: nil)
```
aforge: Remove this.

Tinodios/LoginViewController.swift
```diff
@@ -59,7 +59,12 @@ class LoginViewController: UIViewController {
             let storyboard = UIStoryboard(name: "Main", bundle: nil)
             let destinationVC = storyboard.instantiateViewController(withIdentifier: "ChatsNavigator") as! UINavigationController
 
-            self.show(destinationVC, sender: nil)
+            if let window = UIApplication.shared.keyWindow {
```
aforge: Interesting!
It turns out it's the recommended way to get |window| for Swift 4.2.
Can you also update UiUtils.routeToLoginVC to also follow this convention please?

or-else: done

Tinodios/LoginViewController.swift
or-else: done

Tinodios/FindInteractor.swift
or-else: done

PR #42, opened 2019-07-22
Copy and delete messages

or-else:
Copy and Delete actions added to MessageVC on long tap. Crash on `{pres what=tags}` fixed. Some other minor updates.


Tinodios/MessageViewController.swift
```diff
@@ -928,14 +934,47 @@ extension MessageViewController : MessageCellDelegate {
...
+        UIMenuController.shared.setMenuVisible(true, animated: true)
+    }
+
+    @objc func copyMessageContent(sender: UIMenuController) {
+        guard let menuItem = sender.menuItems?.first as? MessageMenuItem, menuItem.seqId > 0, let msgIndex = messageSeqIdIndex[menuItem.seqId] else { return }
+
+        let msg = messages[msgIndex]
+
+        var senderName: String?
+        if let sub = topic?.getSubscription(for: msg.from), let pub = sub.pub {
```
aforge: This works for group topics but doesn't for p2p topics. Can you check please?

TinodeSDK/Topic.swift
```diff
@@ -1329,10 +1338,13 @@ open class MeTopic<DP: Codable>: Topic<DP, PrivateType, DP, PrivateType> {
                 let acs = Acs()
                 acs.update(from: pres.dacs)
                 if acs.isModeDefined {
-                    getMeta(query: getMetaGetBuilder().withGetSub(user: pres.src).build());
+                    getMeta(query: getMetaGetBuilder().withSub(user: pres.src).build());
```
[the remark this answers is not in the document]
or-else: Removed here and in a few other places.

Tinodios/MessageViewController.swift
or-else: Works fine:

> [Dave Goliathsson]: Convictions cause convicts -- what you believe imprisons you.; 6/9/19, 4:00 AM09

aforge: Humm... Strange. I've just pulled this PR locally and tried it again.
It's unstable for some reason: it works for some topics and doesn't work for others.
> [Unknown usrui....]: Cowards die many times before

It looks like some of these subscriptions are missing |pub| for some reason.
Anyway, let this not block us.

or-else: Which one is that?

aforge: I logged in as Alice and opened a topic Bob:
- copying messages from Bob works
- copying messages from Alice (self) doesn't

Then I opened a topic with Eve: copying messages from Eve doesn't work.
I repeated the same exercise with Dave: copying messages from Dave doesn't work.
Is it working for you for all topics?

or-else: Yes, I can reproduce it. I'll look into it.

PR #44, opened 2019-07-23
Bug fixes

or-else:
* Show correct context menu when input field contains text.
* Remove wrong menu items when the context is the input field.

iOS is weird.


Tinodios/MessageViewController.swift
```diff
@@ -456,7 +456,7 @@ extension MessageViewController: MessageDisplayLogic {
             // Ensure uniqueness of values. No need to reload newly inserted values.
             // The app will crash if the same index is marked as removed and refreshed. Which seems
             // to be an Apple bug because removed index is against the old array, refreshed against the new.
-            refresh = Array(Set(refresh).subtracting(Set(diff.inserted)).subtracting(Set(diff.removed)))
+            refresh = Array(Set(refresh).subtracting(Set(diff.inserted)))
```
aforge: Interesting. It fails when you delete and refresh an item in one go but doesn't when you do it step by step?

Tinodios/MessageViewController.swift
```diff
@@ -945,6 +944,13 @@ extension MessageViewController : MessageCellDelegate {
     }
 
     func createPopupMenu(in cell: MessageCell) {
+        // Make cell the first responder otherwise menu will show wrong items.
+        if sendMessageBar.inputField.isFirstResponder {
```
aforge: Really cool!

Tinodios/MessageViewController.swift
or-else: It crashes it I remove and update the same index although it should not according to logic because delete index is against the old array and update is against the new. These different items.

or-else: Effectively it's impossible to reliad the item which follows deleted item.

PR #45, opened 2019-07-24
fix for missing sender name in some p2p topics

or-else:
Also removed commented out code
Fix for uploaded avatars in TopicInfo and AccountSetting being too large


Tinodios/UiUtils.swift
```diff
@@ -33,7 +33,9 @@ class UiTinodeEventListener : TinodeEventListener {
 
 class UiUtils {
     static let kMinTagLength = 4
-    static let kAvatarSize = 128
+    static let kAvatarSize: CGFloat = 128
+    static let kMaxBitmapSize: CGFloat = 1024
```
aforge: Do we want to go that large?
1024^2x4 = 4M for the largest upload.

or-else: It's compressed as png and much much smaller, usually around 200-300KB. Yes, it would be nice to have an additional check for bitmap weight because a specially crafted bitmap may be much bigger, but that affects just the user who tries to upload such an image. I.e. it's not a security risk.

or-else: Scratch that. `UIImage.pngData()` returns exceedingly large number of bytes.

or-else: I'll do a bit more work on it before merging.

or-else:
I've changed some things in your code, so I'm re-requesting review for the new changes.

Tinodios/UiUtils.swift
```diff
@@ -302,8 +303,25 @@ extension UIViewController {
 }
 
 extension UIImage {
+    private static let kScaleFactor: CGFloat = 0.70710678118 // 1.0/SQRT(2)
+
+    private static func resizeImage(image: UIImage, newSize size: (dst: CGSize, src: CGRect, altered: Bool)) -> UIImage? {
```
aforge: Please, declare (cst: CGSize, src: CGRect, altered: Bool) as a typealias.

Tinodios/MessageViewController+SendMessageBarDelegate.swift
```diff
@@ -78,41 +78,17 @@ extension MessageViewController : UIDocumentPickerDelegate {
...
-            bits = resizedBits
-        }
-        let width = Int(image.size.width)
-        let height = Int(image.size.height)
-        let mimeType = "image/png"
-        let fname = "fn.png"
+        // Ensure image size in bytes and liner dimensions are under the limits.
+        guard let image = image?.resize(byteSize: MessageViewController.kMaxInbandAttachmentSize, asMimeType: mimeType)?.resize(width: UiUtils.kMaxBitmapSize, height: UiUtils.kMaxBitmapSize, clip: false) else { return }
+
+        guard let bits = mimeType != "image/png" ? image.jpegData(compressionQuality: 0.8) : image.pngData() else { return }
```
aforge: This piece of code:
mimeType != "image/png" ? image.jpegData(compressionQuality: 0.8) : image.pngData()
repeats in multiple places.
How about making it a small utility function? Or a UIImage extension? Like getData(for mimeType: String) -> Data?

Tinodios/AttributedStringFormatter.swift
```diff
@@ -414,7 +414,7 @@ class AttributedStringFormatter: DraftyFormatter {
             case .image(let image):
                 let wrapper = NSTextAttachment()
                 wrapper.image = image
-                let (scaledSize, _) = image.sizeUnder(maxWidth: size.width, maxHeight: size.height, clip: false)
+                let (scaledSize, _, _) = image.sizeUnder(maxWidth: size.width, maxHeight: size.height, clip: false)
```
aforge: I think you can change it to
let scaledSize = image.sizeUnder(maxWidth: size.width, maxHeight: size.height, clip: false).dst

Tinodios/UiUtils.swift
or-else: done

Tinodios/MessageViewController+SendMessageBarDelegate.swift
or-else: done

Tinodios/AttributedStringFormatter.swift
or-else: done

PR #48, opened 2019-07-26
Bug fixes

or-else:


Tinodios/UiUtils.swift
```diff
@@ -21,7 +21,9 @@ class UiTinodeEventListener : TinodeEventListener {
     }
     func onDisconnect(byServer: Bool, code: Int, reason: String) {
         connected = false
-        UiUtils.showToast(message: "Connection to server lost.")
+        DispatchQueue.main.async {
```
aforge: In order to prevent such issues from happening, it'd be good to run the function body on the UI thread instead of calling the function.
What do you think?

or-else: Sounds like a good idea. I don't think there is much overhead if it's already called on the main thread.

PR #50, opened 2019-07-27
Corner case permissions

or-else:
* "Not available" toggles in real time on changes to users "W" permission
* "Peer's messaging disabled. Enable" toggles in real time on peer's "W" permission changes


Tinodios/widgets/PermissionsEditViewController.swift
```diff
@@ -75,13 +75,13 @@ class PermissionsEditViewController: UIViewController {
...
             for p in disablePermissions {
                 switch p {
-                case .join: self.joinSwitch.isEnabled = false
-                case .read: self.readSwitch.isEnabled = false
-                case .write: self.writeSwitch.isEnabled = false
-                case .notifications: self.notificationsSwitch.isEnabled = false
-                case .approve: self.approveSwitch.isEnabled = false
-                case .invite: self.inviteSwitch.isEnabled = false
-                case .delete: self.deleteSwitch.isEnabled = false
+                case .join: self.joinSwitch.isHidden = true
```
aforge: Yeah, I think we'll need to modify this dialog box code to remove the missing rows and shrink it to only display the available ones.
It looks pretty weird to see these rows without the switches :-)
But perhaps not at the moment.

Tinodios/MessageInteractor.swift
```diff
@@ -366,6 +366,9 @@ class MessageInteractor: DefaultComTopic.Listener, MessageBusinessLogic, Message
     override func onData(data: MsgServerData?) {
         self.loadMessages()
     }
+    override func onPres(pres: MsgServerPres) {
+        DispatchQueue.main.async { self.presenter?.applyTopicPermissions() }
```
aforge: Actually, applyTopicPermissions runs on the UI thread.
So DispatchQueue.main.async {...} is redundant here.
(In general, I think presenter is the right place to switch execution to the UI thread).

Tinodios/widgets/PermissionsEditViewController.swift
or-else: Sorry, that was supposed to go to the next PR. Let me finish it then.

Tinodios/MessageInteractor.swift
or-else: Then this should apply to the rest of callbacks, like onOnline and onInfo?

aforge: Correct.
Note that onOnline calls presenter.setOnline which, in turn, invokes MessageVC.setOnline. But the latter doesn't get called on the UI thread.
As for the other functions, we need to look at it on the case-by-case basis.
If you don't have objections or concerns, I'm going to make the changes.

Tinodios/widgets/PermissionsEditViewController.swift
aforge: Sure, no worries!

Tinodios/MessageInteractor.swift
or-else: OK

Tinodios/widgets/PermissionsEditViewController.swift
or-else: Done. Please review

Tinodios/MessageViewController.swift
```diff
@@ -1070,10 +1067,12 @@ extension MessageViewController : MessageCellDelegate {
         // TODO: move logic to MessageInteractor.
         guard let data = MessageViewController.extractAttachment(from: cell), !data.isEmpty else { return }
         let d = data[0]
-        guard let filename = url.extractQueryParam(withName: "filename") else { return }
+        // FIXME: use actual mime instead of nil when generating file name.
+        let filename = url.extractQueryParam(withName: "filename") ?? Utils.uniqueFilename(forMime: nil)
         let documentsUrl: URL = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first!
         let destinationURL = documentsUrl.appendingPathComponent(filename)
         do {
+            try FileManager.default.createDirectory(at: documentsUrl, withIntermediateDirectories: true, attributes: nil)
```
aforge: Why is this necessary?
Can documentDirectory be missing?

Tinodios/widgets/PermissionsEditViewController.swift
```diff
@@ -7,108 +7,110 @@
...
-            self.onChange?((self.joinSwitch.isOn,
-                            self.readSwitch.isOn,
-                            self.writeSwitch.isOn,
-                            self.notificationsSwitch.isOn,
-                            self.approveSwitch.isOn,
-                            self.inviteSwitch.isOn,
-                            self.deleteSwitch.isOn))
-        }
-        self.dismiss(animated: false, completion: nil)
+        // Canculate added, removed. If there is a difference, call onChange.
```
aforge: s/Canculate/Calculate
Also, add this as a TODO?

Tinodios/widgets/PermissionsEditViewController.swift
```diff
@@ -7,108 +7,110 @@
...
-        approve: Bool, invite: Bool, delete: Bool)
-    public typealias OnChangeHandler = ((
-        _ permissions: PermissionsTuple) -> ())
+class PermissionsEditViewController: UIViewController, UITableViewDelegate, UITableViewDataSource {
+    private static let kAllPermissions: [Character] = ["J","R","W","P","A","S","D","O"]
+    private static let kPermissions: [Character : String] = ["J" : "Join", "R" : "Read", "W" : "Write", "P" : "Receive notifications", "A" : "Approve new members", "S" : "Invite new members", "D" : "Delete messages", "O" : "Full control (owner)"]
+
+    public typealias ChangeHandler = ((_ permissions: String) -> ())
 
+    private static let buttonBorderColor = UIColor(red: 224.0/255.0, green: 224.0/255.0, blue: 224.0/255.0, alpha: 1)
```
aforge: Optional: you can also UIColor(fromHexCode) (https://github.com/tinode/ios/blob/devel/Tinodios/UiUtils.swift#L457)

Tinodios/widgets/PermissionsEditViewController.swift
```diff
@@ -7,108 +7,110 @@
...
-class PermissionsEditViewController: UIViewController {
-    public enum PermissionType {
-        case join, read, write, notifications, approve, invite, delete
-    }
-    public typealias PermissionsTuple = (
-        join: Bool, read: Bool, write: Bool, notifications: Bool,
-        approve: Bool, invite: Bool, delete: Bool)
-    public typealias OnChangeHandler = ((
-        _ permissions: PermissionsTuple) -> ())
+class PermissionsEditViewController: UIViewController, UITableViewDelegate, UITableViewDataSource {
```
aforge: That's an interesting approach.
I was thinking about something different - like removing disabled rows and changing the layout constraints - but the TableView also works :-)

Tinodios/Utils.swift
```diff
@@ -94,6 +95,16 @@ class Utils {
         let candidates = tagsString.split(separator: ",")
         return candidates.filter { $0.count >= UiUtils.kMinTagLength }.map { String($0) }
     }
+
+    public static func uniqueFilename(forMime mime: String?) -> String {
```
aforge: The argument is unused. Is it expected?
Or should it be "mime ?? mimeType" on line 103?

Tinodios/MessageViewController.swift
or-else: `~/Pictures` did not exist and I believe it's the same with `~/Documents` unless Junky Junk was run first. I'll double check, but there is no harm in ensuring that the directory exists.

Tinodios/widgets/PermissionsEditViewController.swift
or-else: The table also looks much nicer: switches are very heavy.

Tinodios/widgets/PermissionsEditViewController.swift
or-else: Nice! Thanks.

Tinodios/ImagePreviewController.swift
```diff
@@ -56,7 +53,21 @@ class ImagePreviewController : UIViewController, UIScrollViewDelegate {
...
+    @IBAction func saveImageButtonClicked(_ sender: Any) {
+        guard let content = previewContent, let imageBits = content.imageBits else { return }
+
+        let picturesUrl: URL = FileManager.default.urls(for: .picturesDirectory, in: .userDomainMask).first!
+        let destinationURL = picturesUrl.appendingPathComponent(content.fileName ?? Utils.uniqueFilename(forMime: content.contentType))
+        do {
+            try FileManager.default.createDirectory(at: picturesUrl, withIntermediateDirectories: true, attributes: nil)
+            try imageBits.write(to: destinationURL)
+            UiUtils.presentFileSharingVC(for: destinationURL)
+        } catch let err as NSError {
```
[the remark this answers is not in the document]
or-else: Did not know that. Thanks. Fixed.

Tinodios/Utils.swift
or-else: Fixed

Tinodios/widgets/PermissionsEditViewController.swift
or-else: Yes. Done.

Tinodios/widgets/PermissionsEditViewController.swift
aforge: Agreed. The table's looking pretty cool.

PR #51, opened 2019-07-31
A bit of clean up around error handling

or-else:
* Warning and Info toasts
* `TinodeError` complied with `LocalizedError` to create useful `.localizedDescription`.
* Make sure `PromisedReply.thenFinally` is really final.


Tinodios/MessageView.swift
```diff
@@ -106,16 +106,17 @@ extension MessageView {
...
+        let blurEffectView = UIVisualEffectView(effect: UIBlurEffect(style: .light))
+        blurEffectView.alpha = 1
         blurEffectView.translatesAutoresizingMaskIntoConstraints = false
         self.addSubview(blurEffectView)
+
         // Pin the edges to the superview edges.
         if #available(iOS 11.0, *) {
             NSLayoutConstraint.activate([
-                blurEffectView.topAnchor.constraint(equalTo: self.safeAreaLayoutGuide.topAnchor),
+                blurEffectView.topAnchor.constraint(equalTo: self.topAnchor),
```
aforge: Why change this?

or-else: When there are messages in the topic and the title bar is minimized, the `self.safeAreaLayoutGuide.topAnchor` leaves a gap at the top the size of the difference between the normal title bar and the minimized one.

PR #53, opened 2019-08-01
Bug fixes

or-else:
* Promises refactored a bit to correctly handle the case of `promise.thenCatch(...).thenApply(...)`: if `onFailure` returns `nil`, call the subsequent `thenApply` with a `nil` argument instead of throwing `PromisedReplyError.illegalStateError("called handleSuccess on a non-resolved promise")`.
* Work around for an apple bug when deleting the last message in a conversation.
* Bug fix for access mode being changed even if the server rejects the change.

### DIFF
--- a/TinodeSDK/PromisedReply.swift
+++ b/TinodeSDK/PromisedReply.swift
@@ -46,12 +46,12 @@ private class CountDownLatch {
 }
 
 public class PromisedReply<Value> {
-    public typealias SuccessHandler = ((Value) throws -> PromisedReply<Value>?)?
+    public typealias SuccessHandler = ((Value?) throws -> PromisedReply<Value>?)?
     public typealias FailureHandler = ((Error) throws -> PromisedReply<Value>?)?
     public typealias FinallyHandler = (() throws -> PromisedReply<Value>?)
     enum State {
         case waiting
-        case resolved(Value)
+        case resolved(Value?)
         case rejected(Error)
         
         var isDone: Bool {
@@ -103,7 +103,7 @@ public class PromisedReply<Value> {
         countDownLatch = CountDownLatch(count: 0)
     }
 
-    func resolve(result: Value) throws {
+    func resolve(result: Value?) throws {
         defer {
             countDownLatch?.countDown()
             // down the semaphore
@@ -112,7 +112,7 @@ public class PromisedReply<Value> {
         try queue.sync {
             // critical section
             guard case .waiting = state else {
-                throw PromisedReplyError.illegalStateError("Resolve: Promise already resolved.")
+                throw PromisedReplyError.illegalStateError("Resolve: Promise already completed.")
             }
             state = .resolved(result)
             try callOnSuccess(result: result)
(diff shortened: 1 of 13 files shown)


TinodeSDK/PromisedReply.swift
```diff
@@ -46,12 +46,12 @@ private class CountDownLatch {
 }
 
 public class PromisedReply<Value> {
-    public typealias SuccessHandler = ((Value) throws -> PromisedReply<Value>?)?
+    public typealias SuccessHandler = ((Value?) throws -> PromisedReply<Value>?)?
```
aforge: I'm also wondering (unrelated to this change): do we ever throw exceptions in the handlers that we don't handle in their bodies?
That is, do we need "throws" here?

TinodeSDK/Tinode.swift
```diff
@@ -485,7 +484,7 @@ public class Tinode {
                             lang: kLocale))
         return try! sendWithPromise(payload: msg, with: msgId).thenApply(
             onSuccess: { [weak self] pkt in
-                guard let ctrl = pkt.ctrl else {
+                guard let pkt = pkt, let ctrl = pkt.ctrl else {
```
aforge: guard let ctrl = pkt?.ctrl else { ... }

TinodeSDK/Tinode.swift
```diff
@@ -834,13 +833,13 @@ public class Tinode {
...
                         try connected.resolve(result: pkt)
                     }
-                    let ctrl = pkt.ctrl!
-                    tinode.timeAdjustment = Date().timeIntervalSince(ctrl.ts)
-                    // tinode store
-                    tinode.store?.setTimeAdjustment(adjustment: tinode.timeAdjustment)
-                    // listener
-                    tinode.listener?.onConnect(
-                        code: ctrl.code, reason: ctrl.text, params: ctrl.params)
+                    if let pkt = pkt, let ctrl = pkt.ctrl {
```
aforge: ditto

TinodeSDK/Topic.swift
```diff
@@ -530,7 +530,7 @@ open class Topic<DP: Codable, DR: Codable, SP: Codable, SR: Codable>: TopicProto
                 let isAttached = self?.attached ?? false
                 if !isAttached {
                     self?.attached = true
-                    if let ctrl = msg.ctrl {
+                    if let msg = msg, let ctrl = msg.ctrl {
```
aforge: ditto

TinodeSDK/PromisedReply.swift
or-else: I think it's useful Here is a use case (pseudo code):
```
promise1.thenApply({
   promise2.thenApply() // this one fails, but has no catcher.
                        // So it throws to be caught in the failureHandler.
}).thenCatch(failureHandler)
```

TinodeSDK/Topic.swift
or-else: done

TinodeSDK/Tinode.swift
or-else: done

TinodeSDK/Tinode.swift
or-else: done

PR #55, opened 2019-08-03
More iOS-like behavior of AccountSettings and TopicInfo forms

or-else:
* Got rid of buttons in forms, table rows are now using standard style whenever possible. 
* Added "No access to messages" to blocked topic + subtitle with an explanation.


Tinodios/MessageViewController.swift
```diff
@@ -15,7 +15,7 @@ protocol MessageDisplayLogic: class {
     func displayChatMessages(messages: [StoredMessage])
     func reloadMessage(withSeqId seqId: Int)
     func updateProcess(forMsgId msgId: Int64, progress: Float)
-    func applyTopicPermissions()
+    func applyTopicPermissions(withMessage: String?)
```
aforge: withErrorMessage perhaps?
To make it clear there was an error and we are restricting permissions.

or-else: Let's make it `withError: Error?` then.

PR #57, opened 2019-08-04
Visual fixes, spinners

or-else:
* Changed the tags icon to be more like other icons (glyph vs outline).
* Fixed tint not applied to some icons.
* Show progress spinner in LoginVC.
* Add missing `super` calls for some overriden methods.


Tinodios/UiUtils.swift
```diff
@@ -311,6 +311,22 @@ class UiUtils {
...
+
+            let loadingIndicator = UIActivityIndicatorView(frame: CGRect(x: 10, y: 5, width: 50, height: 50))
+            loadingIndicator.hidesWhenStopped = true
+            loadingIndicator.style = UIActivityIndicatorView.Style.gray
+            loadingIndicator.startAnimating();
+
+            alert.view.addSubview(loadingIndicator)
+            parent.present(alert, animated: true, completion: nil)
+        } else if let vc = parent.presentedViewController, vc is UIAlertController {
+            parent.dismiss(animated: false, completion: nil)
```
aforge: Should dismiss be called on vc and not parent?

Tinodios/UiUtils.swift
```diff
@@ -311,6 +311,22 @@ class UiUtils {
...
     }
+
+    public static func toggleProgressOverlay(in parent: UIViewController, visible: Bool, title: String? = nil) {
+        if visible {
+            let alert = UIAlertController(title: nil, message: title ?? "Please wait...", preferredStyle: .alert)
+
+            let loadingIndicator = UIActivityIndicatorView(frame: CGRect(x: 10, y: 5, width: 50, height: 50))
+            loadingIndicator.hidesWhenStopped = true
+            loadingIndicator.style = UIActivityIndicatorView.Style.gray
+            loadingIndicator.startAnimating();
```
aforge: Remove semicolon please.

Tinodios/UiUtils.swift
or-else: Calling on the parent is the way to go:
https://developer.apple.com/documentation/uikit/uiviewcontroller/1621505-dismiss
> `dismiss(animated:completion:)`
> Dismisses the view controller that was presented modally by the view controller.

It works as expected. Try entering a wrong password.

It **can** be called on the alert VC though:
> If you call this method on the presented view controller itself, UIKit asks the presenting view controller to handle the dismissal.

Tinodios/UiUtils.swift
or-else: done here and in a few other places.

Tinodios/UiUtils.swift
aforge: Nice! Thanks!

PR #60, opened 2019-08-05
Sign up works on all screens.

or-else:
Completely refactored Sign Up screen to work on all device sizes and in both landscape and portrait mode.

Error reporting for blank UITextFields does not look good with the new design. I'll update it later.


Tinodios/SignupViewController.swift
```diff
@@ -63,16 +61,16 @@ class RegisterViewController: UIViewController {
...
             return
         }
-        
-        signUpBtn.isUserInteractionEnabled = false
+
+        signUpButton.isUserInteractionEnabled = false
         let tinode = Cache.getTinode()
 
-        let avatar = uploadedAvatar ? avatarView?.image?.resize(width: 128, height: 128, clip: true) : nil
+        let avatar = uploadedAvatar ? avatarImageView?.image?.resize(width: 128, height: 128, clip: true) : nil
```
aforge: Please, use UiUtils.kAvatarSize instead of the 128 literals.

Tinodios/SignupViewController.swift
```diff
@@ -115,31 +113,32 @@ class RegisterViewController: UIViewController {
...
-                })?.thenFinally(finally: { [weak self] in
-                    DispatchQueue.main.async {
-                        self?.signUpBtn.isUserInteractionEnabled = true
-                    }
-                    return nil
-                })
-            
+            })?.thenFinally(finally: { [weak self] in
+                DispatchQueue.main.async {
+                    self?.signUpButton.isUserInteractionEnabled = true
```
aforge: Can you add UiUtils.toggleProgressOverlay() here as well please?

Tinodios/SignupViewController.swift
or-else: done

Tinodios/SignupViewController.swift
or-else: done

PR #61, opened 2019-08-07
Group member editing refactored

or-else:
* EditMembers VC refactored into a component.
* New Group VC refactored to inherit from UITableView  to be useable on small screens.
* New Group & Topic Info VCs changed to use the new unified EditMembers VC.
* More IBDesignable elements added to TagsEditView. It can be now styled from the storyboard to be like other UITextFields (needs more work)
* Bug fixes.
* Use red icon `( ! )` instead of red border when a required UITextField is empty.


Tinodios/EditMembersViewController.swift
```diff
@@ -92,64 +96,73 @@ class EditMembersViewController: UIViewController, UITableViewDataSource {
...
+        guard let uid = contacts[indexPath.row].uniqueId else {
+            print("no unique id for user at \(indexPath.row - 1)")
+            return indexPath
+        }
+
+        return delegate?.editMembersWillChangeState(editMembersView, uid: uid, added: false, initiallySelected: initialIds.contains(uid)) ?? true ? indexPath : nil
+    }
+
     func tableView(_ tableView: UITableView, didDeselectRowAt indexPath: IndexPath) {
+        guard let uid = contacts[indexPath.row - 1].uniqueId else {
```
aforge: Why indexPath.row - 1 here? Is it intended?

Tinodios/NewGroupViewController.swift
```diff
@@ -123,77 +128,39 @@ class NewGroupViewController: UIViewController, UITableViewDataSource {
...
+        return selectedContacts.compactMap { $0.uniqueId }
     }
-}
 
-extension NewGroupViewController : UICollectionViewDataSource {
-    func collectionView(_ collectionView: UICollectionView, numberOfItemsInSection section: Int) -> Int {
-        toggleNoSelectedMembersNote(on: selectedContacts.isEmpty)
-        return selectedContacts.count
+    func editMembersDidEndEditing(_: UIView, added: [String], removed: [String]) {
+        self.interactor?.updateSelection(added: added, removed: removed)
```
aforge: This updates NewGroupInteractor.selectedContacts but doesn't NewGroupViewController.selectedContacts. Because of this, when you add some members to a new group, they don't show up in the table.
So we'll need to update the latter.
In general, it appears that apart from NewGroupInteractor.createGroupTopic, there's nothing more than the selectedContacts logic.
Should we just move the logic to NewGroupViewController altogether and remove NewGroupInteractor?

Tinodios/EditMembersViewController.swift
or-else: Yep, it's a bug due to a copy-paste fro NewGroup. Fixed.

Tinodios/NewGroupViewController.swift
or-else: OK, will do it now.

or-else: Done. I also had to update UserDb: there were some unused leftovers from android + I added loading users by a list of IDs.

Tinodios/account/ContactsManager.swift
```diff
@@ -60,16 +60,24 @@ class ContactsManager {
...
-                uniqueId: q.uid)
+    public func fetchContacts(withUids uids: [String]? = nil) -> [ContactHolder]? {
+        let users: [UserProto]?
+        if uids != nil {
+            users = userDb.read(uids: uids!)
+        } else {
+            guard let uid = Cache.getTinode().myUid else { return nil }
+            users = userDb.readAll(for: uid)
+        }
+        if let users = users {
```
aforge: guard let users = users else { return nil } 
is better here to avoid indentation.

or-else: `guard let users2 = users else { return nil }` - I would have to allocate another variable. That's why I decided against `guard`

or-else: I simplified it in a different way.

PR #63, opened 2019-08-09
Big fixes, Reset Password VC

or-else:
* Reset password VC implemented
* fix: Attachment action sheet is not visible when keyboard is up
* fix: Group ownership transfer is not possible
* fix: TopicInfo: [+ Add members] is still shown even when the user has no A or S permission
* remove redundant { get } designators
* make sure tintColor is applied to icons (xcode bug)
* fix: text fields on Login VC need to be focused when the keyboard is shown
* Login VC styling is changed: navbar removed, logo is horizontal now, layout simplified.


TinodeSDK/AuthScheme.swift
```diff
@@ -43,7 +44,14 @@ struct AuthScheme {
         }
         return (uname + ":" + password).toBase64()!
     }
-    
+
+    static func encodeResetToken(scheme: String, method: String, value: String) throws -> String {
+        guard scheme.range(of: ":") == nil && method.range(of: ":") == nil else {
```
aforge: !scheme.contains(":") ?

or-else: done

PR #69, opened 2019-08-19
Scheme enforcement for external URLs

or-else:


Tinodios/AttributedStringFormatter.swift
```diff
@@ -83,6 +83,7 @@ class AttributedStringFormatter: DraftyFormatter {
         case "url":
             guard let ref = attr["ref"]?.asString() else { return nil }
             guard let urlc = URLComponents(string: ref) else { return nil }
+            guard urlc.scheme == "http" || urlc.scheme == "https" else { return nil }
```
aforge: Perhaps, move the logic to Utils.swift?
E.g. acceptUrlScheme(scheme: String) -> Bool

or-else: It was only a couple of places. Did not feel like it was worth a method.

PR #78, opened 2019-10-21
Simplified Chinese translation

KyoShio:
Simplified Chinese translation


or-else:
Wow! Excellent! Thanks a lot!

or-else:
BTW, there is also text in the Settings.bundle:
https://github.com/tinode/ios/blob/master/Tinodios/Settings.bundle/en.lproj/Root.strings

PR #81, opened 2019-10-23
Change Tinode link in invitation message to tinode.co.

aforge:


or-else:
Ah, I missed this one. OK. We probably need to set up some kind of redirect to appropriate store. This will do for now though.

PR #84, opened 2019-10-26
Substitute TinodeSDK.Log.log with NSLog.

aforge:
Also, use correct format strings.


or-else:
The need to use `%@` as format specifier for strings was a side effect of os_log being wrapped. It works but it's less efficient than `%s`. I'll merge now but let's fix it over time.

PR #85, opened 2019-10-30
Add credMethods to StoredAccount.

aforge:
Following the logic of https://github.com/tinode/tindroid/commit/5e13af04bcb7fa62c22bbd9c423a278c88e14004

### DIFF
--- a/TinodeSDK/Tinode.swift
+++ b/TinodeSDK/Tinode.swift
@@ -775,23 +775,10 @@ public class Tinode {
             // Load topics if not yet loaded.
             loadTopics()
         } else {
-            var meth: [String]? = nil
-            if let params = ctrl.params {
-                for (key, value) in params {
-                    if key == "cred", case let .string(m) = value {
-                        if meth == nil {
-                            meth = [String]()
-                        }
-                        meth!.append(m)
-                    }
-                }
-            }
+            let meth = ctrl.getStringArray(for: "cred")
             store?.setMyUid(uid: newUid!, credMethods: meth)
         }
         isConnectionAuthenticated = true
-        if let t = authToken, !autoLogin {
-            setAutoLoginWithToken(token: t)
-        }
         listener?.onLogin(code: ctrl.code, text: ctrl.text)
     }
     private func updateAccountSecret(uid: String?, scheme: String, secret: String) -> PromisedReply<ServerMessage>? {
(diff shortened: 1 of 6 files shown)


TinodeSDK/Tinode.swift
```diff
@@ -768,18 +768,31 @@ public class Tinode {
...
         if ctrl.code < 300 {
-            isConnectionAuthenticated = true
-            if let t = authToken, !autoLogin {
-                setAutoLoginWithToken(token: t)
+            store?.myUid = newUid
+            // Load topics if not yet loaded.
+            loadTopics()
+        } else {
+            var meth: [String]? = nil
+            if let params = ctrl.params {
```
or-else: Why not just 
```
var meth = ctrl.params?.getStringArray(for: "cred")
```

TinodeSDK/Tinode.swift
```diff
@@ -768,18 +768,31 @@ public class Tinode {
...
+                        meth!.append(m)
+                    }
+                }
             }
-            listener?.onLogin(code: ctrl.code, text: ctrl.text)
+            store?.setMyUid(uid: newUid!, credMethods: meth)
+        }
+        isConnectionAuthenticated = true
+        if let t = authToken, !autoLogin {
+            setAutoLoginWithToken(token: t)
```
or-else: Interesting. This is different from Android. On Android the app calls setAutoLoginWithToken. Here the SDK calls it, the app loses ability to control setAutoLoginWithToken. Maybe then it should be made private?

If `autoLogin` is true the token is not saved. Does not seem logical to me.

Tinodios/db/AccountDb.swift
```diff
@@ -56,39 +61,47 @@ public class AccountDb {
...
+                    result = StoredAccount(id: newId, uid: uid, credMethods: meth)
                 }
                 if result!.id < 0 {
                     result = nil
+                } else {
+                    result?.credMethods = meth
                 }
             }
         } catch {
+            print("failed to add account '\(error)'")
```
or-else: `Log.debug....` etc?

TinodeSDK/Tinode.swift
aforge: Ah, okay. This one was copied over from an earlier Android version.
Yes, I agree it should be called from the app.

PR #86, opened 2019-10-31
Connect and log in in the background on app startup.

aforge:


Tinodios/AppDelegate.swift
```diff
@@ -21,31 +21,38 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
                     }
-                    UiUtils.routeToChatListVC()
+                } catch SwiftWebSocket.WebSocketError.network(let e)  {
+                    // No network connection.
+                    Cache.log.debug("AppDelegate [network] - could not connect to Tinode: %@", e)
                     success = true
+                } catch {
+                    Cache.log.error("AppDelegate - failed to automatically login to Tinode: %@", error.localizedDescription)
+                }
+                if !success {
```
or-else: This seems to be too harsh. 5xx error should not lead to log out. And maybe some 4xx errors either (like `409 already authenticated`).

aforge: Yes. Added these two. Other 400s in datamodel.go look irrelevant.

PR #89, opened 2019-11-04
Fixes for AccountInfo and TopicInfo VCs.

aforge:
1. Allow partial pub and priv merge in Description.merge().
2. Deal with UI blinking in TopicInfoVC.
3. Allow multiple event listeners in Tinode.

### DIFF
--- a/TinodeSDK/Tinode.swift
+++ b/TinodeSDK/Tinode.swift
@@ -517,7 +517,7 @@ public class Tinode {
         return topics[topicName] != nil
     }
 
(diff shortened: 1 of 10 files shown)


TinodeSDK/Tinode.swift
```diff
@@ -186,6 +186,71 @@ public class Tinode {
             }
         }
     }
+
+    // Forwards events to all subscribed listeners.
+    private class ListenerNotifier: TinodeEventListener {
+        private var listeners = [TinodeEventListener]()
```
or-else: I think Swift 4+ recommends
`listeners : [TinodeEventListener] = []`

PR #93, opened 2019-11-08
limit maximum length of topic title, private info, and user name

or-else:


Tinodios/AccountSettingsViewController.swift
```diff
@@ -284,7 +284,7 @@ class AccountSettingsViewController: UITableViewController {
         guard let newTitle = newTitle else { return }
         let pub = me.pub == nil ? VCard(fn: nil, avatar: nil as Data?) : me.pub!.copy()
         if pub.fn != newTitle {
-            pub.fn = newTitle
+            pub.fn = String(newTitle.prefix(UiUtils.kMaxTitleLength))
```
aforge: Do we want it as prefix only?
Maybe a prefix and suffix, e.g. "abcdefghijklmnopqrstuvwxyz" -> "abc...xyz"

Tinodios/NewGroupViewController.swift
```diff
@@ -136,15 +136,15 @@ class NewGroupViewController: UITableViewController {
...
+        let groupName = UiUtils.ensureDataInTextField(groupNameTextField, maxLength: UiUtils.kMaxTitleLength)
         let tinode = Cache.getTinode()
         let members = selectedMembers.filter { !tinode.isMe(uid: $0) }
         if members.isEmpty {
             UiUtils.showToast(message: "Select at least one group member")
             return
         }
         // Optional
-        let privateInfo = (privateTextField.text ?? "").trimmingCharacters(in: .whitespacesAndNewlines)
+        let privateInfo = String((privateTextField.text ?? "").trimmingCharacters(in: .whitespacesAndNewlines).prefix(UiUtils.kMaxTitleLength))
```
aforge: I'd move the formatting function to UiUtils so if we want to change it, there's only one place where we need to make changes.

Tinodios/AccountSettingsViewController.swift
or-else: I would rather not modify the title too much. If the user pastes a very long string into title we just cut it short. Otherwise it would be too unexpected.

PR #94, opened 2019-11-08
Making dark mode usable

or-else:
This is work in progress. I'm fixing dark mode. It seems to be pretty straightforward.

### DIFF
--- a/Tinodios/MessageCell.swift
+++ b/Tinodios/MessageCell.swift
@@ -31,14 +31,22 @@ class MessageCell: UICollectionViewCell {
 
     public override init(frame: CGRect) {
         super.init(frame: frame)
-        backgroundColor = UIColor.white
+        if #available(iOS 12.0, *), traitCollection.userInterfaceStyle == .dark {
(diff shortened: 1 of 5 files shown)


Tinodios/MessageCell.swift
```diff
@@ -31,14 +31,22 @@ class MessageCell: UICollectionViewCell {
 
     public override init(frame: CGRect) {
         super.init(frame: frame)
-        backgroundColor = UIColor.white
+        if #available(iOS 12.0, *), traitCollection.userInterfaceStyle == .dark {
```
aforge: Maybe make a UIView extension configuring backgroundColors?

or-else: I don't see how to make that usable. I can wrap `if #available(iOS 12.0, *), traitCollection.userInterfaceStyle == .dark` into an extension but that seems like an overkill.

or-else: There are not that many places where this check is needed. All are done now.

PR #96, opened 2019-11-12
Fixes for missing login and layout

aforge:
Set auto login in SignupVC and CredentialsVC.
Fix message height calculation in MessageVC.
Route back to login screen in reset password VC.

### DIFF
--- a/Tinodios/CredentialsViewController.swift
+++ b/Tinodios/CredentialsViewController.swift
@@ -64,6 +64,9 @@ class CredentialsViewController : UIViewController {
                             UiUtils.showToast(message: "Verification failure: \(ctrl.code) \(ctrl.text)")
                         }
                     } else {
+                        if let token = tinode.authToken {
+                            tinode.setAutoLoginWithToken(token: token)
+                        }
                         UiUtils.routeToChatListVC()
                     }
                     return nil
--- a/Tinodios/MessageViewController.swift
+++ b/Tinodios/MessageViewController.swift
@@ -964,7 +964,16 @@ extension MessageViewController : MessageViewLayoutDelegate {
             attributedText.append(NSAttributedString(string: "none", attributes: [.font: Constants.kContentFont]))
         }
         attributedText.append(NSAttributedString(string: carveout, attributes: [.font: Constants.kContentFont]))
-        let size = attributedText.boundingRect(with: CGSize(width: maxWidth, height: .greatestFiniteMagnitude), options: [.usesLineFragmentOrigin, .usesFontLeading], context: nil).integral.size
+
+        // Calculate content size.
+        // Per https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/TextLayout/Tasks/StringHeight.html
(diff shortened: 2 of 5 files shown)


Tinodios/CredentialsViewController.swift
```diff
@@ -64,6 +64,9 @@ class CredentialsViewController : UIViewController {
                             UiUtils.showToast(message: "Verification failure: \(ctrl.code) \(ctrl.text)")
                         }
                     } else {
+                        if let token = tinode.authToken {
+                            tinode.setAutoLoginWithToken(token: token)
```
or-else: Why was it removed?

Tinodios/MessageViewController.swift
```diff
@@ -964,7 +964,16 @@ extension MessageViewController : MessageViewLayoutDelegate {
             attributedText.append(NSAttributedString(string: "none", attributes: [.font: Constants.kContentFont]))
         }
         attributedText.append(NSAttributedString(string: carveout, attributes: [.font: Constants.kContentFont]))
-        let size = attributedText.boundingRect(with: CGSize(width: maxWidth, height: .greatestFiniteMagnitude), options: [.usesLineFragmentOrigin, .usesFontLeading], context: nil).integral.size
+
+        // Calculate content size.
+        // Per https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/TextLayout/Tasks/StringHeight.html
+        let ts = NSTextStorage(attributedString: attributedText)
+        let tc = NSTextContainer(size: CGSize(width: maxWidth, height: .greatestFiniteMagnitude))
```
or-else: Can it be reused?

Tinodios/MessageViewController.swift
```diff
@@ -964,7 +964,16 @@ extension MessageViewController : MessageViewLayoutDelegate {
...
+        let lm = NSLayoutManager()
```
or-else: ditto - can it be reused?

Tinodios/MessageViewController.swift
```diff
@@ -964,7 +964,16 @@ extension MessageViewController : MessageViewLayoutDelegate {
...
+        lm.addTextContainer(tc)
+        ts.addLayoutManager(lm)
+
+        let size = lm.usedRect(for: tc).integral.size
```
or-else: why allocate size? Why not just `return`?

Tinodios/MessageViewController.swift
aforge: Refactored in to UiUtils.TextSizeHelper.

Tinodios/MessageViewController.swift
aforge: Yes.

Tinodios/MessageViewController.swift

Tinodios/CredentialsViewController.swift
aforge: We never set it. Looking at the existing code, auto login gets set separately in the app.
Both iOS and Android. We may move it to the SDK, though (loginSuccessful).

PR #99, opened 2019-11-13
Fix UI issues in iOS 10.

aforge:
UI problems similar to iOS 9.


Tinodios/FindViewController.swift
```diff
@@ -67,7 +67,7 @@ class FindViewController: UITableViewController, FindDisplayLogic {
         // Monitor when the search button is tapped.
         searchController.searchBar.delegate = self
         self.definesPresentationContext = true
-        if #available(iOS 10.0, *) {
+        if #available(iOS 11.0, *) {
```
or-else: People suggest this format: 
`if #available(iOS 11, *) {} else {`

aforge: Done.

PR #100, opened 2019-11-14
Handle push notifications in Tinodios

aforge:
1. Route to the appropriate MessageVC on tap on a push notification.
2. Process remote notifications and zero the icon badge count.
3. Clear Firebase token on logout.

### DIFF
--- a/Tinodios/AppDelegate.swift
+++ b/Tinodios/AppDelegate.swift
@@ -99,7 +99,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
 
     func applicationDidBecomeActive(_ application: UIApplication) {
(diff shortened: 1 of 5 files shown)


Tinodios/AppDelegate.swift
```diff
@@ -99,7 +99,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
 
     func applicationDidBecomeActive(_ application: UIApplication) {
-        // Restart any tasks that were paused (or not yet started) while the application was inactive. If the application was previously in the background, optionally refresh the user interface.
+        application.applicationIconBadgeNumber = 0
```
or-else: This looks a bit extreme. It would make the counter disappear, but the app would still have unread messages. Something like the logic in webapp would probably be better: update the counter when some messages are read.

aforge: Updated.

PR #101, opened 2019-11-17
Show proper title in arriving notifications.

aforge:
Add TinodiosNSExtension which modifies the notification body.
Move Tinodios/db (sqlite helper code) to TinodiosDB framework
so that the code can be shared between the app and the extension.


Tinodios/Info.plist
```diff
@@ -9,7 +9,7 @@
 	<key>CFBundleExecutable</key>
 	<string>$(EXECUTABLE_NAME)</string>
 	<key>CFBundleIdentifier</key>
-	<string>$(APP_BUNDLE_ID)</string>
+	<string>$(PRODUCT_BUNDLE_IDENTIFIER)</string>
```
or-else: Are you sure of this change? I think it should stay as APP_BUNDLE_ID.

TinodiosNSExtension/NotificationService.swift
```diff
@@ -0,0 +1,57 @@
...
+        if let bestAttemptContent = bestAttemptContent {
+            defer { self.contentHandler!(bestAttemptContent) }
+
+            let store = BaseDb.getInstance().sqlStore!
+
+            let userInfo = bestAttemptContent.userInfo
+            guard let topic = userInfo["topic"] as? String, !topic.isEmpty,
+                let xfrom = userInfo["xfrom"] as? String, !xfrom.isEmpty else { return }
+
+            guard let user = store.userGet(uid: xfrom) as? DefaultUser else { return }
```
or-else: That's actually possible and likely to receive a notification of a message from an unknown user. Maybe we should show the notification anyway saying "unknown"?

or-else: It may happen for example when you are subscribed to a group topic and someone joins it and messages the group while you are not online in the group.

Tinodios/Info.plist
or-else: This is defined in https://github.com/tinode/ios/blob/devel/devel.xcconfig and https://github.com/tinode/ios/blob/devel/prod.xcconfig

TinodiosNSExtension/NotificationService.swift
or-else: The other option is to send a request to server to fetch the description of the user.

Tinodios/Info.plist
or-else: If we use the same bundle ID for test and prod then we can use PRODUCT_BUNDLE_IDENTIFIER.

TinodiosNSExtension/NotificationService.swift
aforge: Actually, this change was not intentional. XCode's been making it automatically lately and I don't know the reason.
Reverted back.

Tinodios/Info.plist
aforge: Actually, this change was not intentional. XCode's been making it automatically lately and I don't know the reason.
Reverted back.

PR #103, opened 2019-11-19
Fix for iOS 10 and images

aforge:


Tinodios/Utils.swift
```diff
@@ -263,4 +263,10 @@ extension StoredMessage {
         cachedContent = AttributedStringFormatter.toAttributed(content, fitIn: size, withDefaultAttributes: attributes)
         return cachedContent
     }
+
+    // Returns true if message contains an inline image.
+    var isImage: Bool {
+        guard let entity = self.content?.entities?[0] else { return false }
+        return entity.data?["val"] != nil && entity.data?["mime"] != nil
```
or-else: You can explicitly check for image by checking entity type

PR #108, opened 2019-11-23
Fix for broken image picker

or-else:
I made image picker uneditable. I.e. it just picks and sends an image without any intermediate steps. Apple's editable image picker is forever broken. There is no way to send an original image if it's in portrait orientation. Everybody seemingly follows these steps:
1. Pick an image from camera or roll with allowsEditing = false
2. Uses custom cropper/editor

We will have to build 2.

I also made Toast taller.

### DIFF
--- a/Tinodios/Info.plist
+++ b/Tinodios/Info.plist
@@ -31,7 +31,7 @@
 	<key>NSPhotoLibraryUsageDescription</key>
 	<string>$(APP_NAME) needs access to your photos to let you you send them to your contacts</string>
 	<key>NSCameraUsageDescription</key>
-	<string>$(APP_NAME) allows you to take pictures with camera and upload these as avatar or messages </string>
+	<string>$(APP_NAME) takes photos with the camera to use as avatar or to send them to your contacts </string>
 	<key>UILaunchStoryboardName</key>
 	<string>LaunchScreen</string>
 	<key>UIMainStoryboardFile</key>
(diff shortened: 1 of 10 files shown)


or-else:
Maybe we can use this: https://github.com/TimOliver/TOCropViewController
It looks to be high-quality + compatible license.

Tinodios/Info.plist
```diff
@@ -31,7 +31,7 @@
 	<key>NSPhotoLibraryUsageDescription</key>
 	<string>$(APP_NAME) needs access to your photos to let you you send them to your contacts</string>
 	<key>NSCameraUsageDescription</key>
-	<string>$(APP_NAME) allows you to take pictures with camera and upload these as avatar or messages </string>
+	<string>$(APP_NAME) takes photos with the camera to use as avatar or to send them to your contacts </string>
```
aforge: It looks like there's a space between "contacts" and "</string>". Remove?

Tinodios/MessageViewController+SendMessageBarDelegate.swift
```diff
@@ -82,7 +82,7 @@ extension MessageViewController : ImagePickerDelegate {
         let mimeType: String = mime == "image/png" ?  mime! : "image/jpeg"
 
         // Ensure image size in bytes and liner dimensions are under the limits.
```
aforge: s/liner/linear

Tinodios/Info.plist
or-else: It was not visible in the UI. Removed

Tinodios/MessageViewController+SendMessageBarDelegate.swift
or-else: done

PR #109, opened 2019-11-24
Keyboard handling fix 

nikletsopen:
This pull request have following purposes: 

1. Fix the issue when collectionView will scroll all the way down when keyboard appears by calculating correct value to add to `collectionView.contentOffset.y` on keyboard appearance. 
2. Make the code more compact and clear, by adding a separate util struct to get all necessary info about the keyboard and its frame. 
3. Remove unnecessary code like `removeKeyboardObservers` - it is not needed anymore since iOS9.

### DIFF
--- a/Tinodios/KeyboardInfo.swift
+++ b/Tinodios/KeyboardInfo.swift
@@ -9,13 +9,13 @@
 import UIKit
 
 struct KeyboardInfo {
-    
+
     let frameBegin: CGRect
     let frameEnd: CGRect
     let isLocal: Bool
     let animationDuration: TimeInterval
     let animationCurve: UIView.AnimationCurve
-    
+
     init?(notification: Notification) {
         guard
             let userInfo = notification.userInfo,
@@ -26,14 +26,14 @@ struct KeyboardInfo {
             let rawCurve = userInfo[UIResponder.keyboardAnimationCurveUserInfoKey] as? Int,
             let animationCurve = UIView.AnimationCurve(rawValue: rawCurve)
             else { return nil }
-        
+
         self.frameBegin = frameBegin
         self.frameEnd = frameEnd
         self.isLocal = isLocal
         self.animationDuration = animationDuration
         self.animationCurve = animationCurve
     }
-    
+
     var animationOptions: UIView.AnimationOptions {
         let option = animationCurve.rawValue << 16
         return UIView.AnimationOptions(rawValue: UInt(option))
--- a/Tinodios/KeyboardInfo.swift
+++ b/Tinodios/KeyboardInfo.swift
@@ -1,0 +1,41 @@
+//
+//  KeyboardInfo.swift
+//  Tinodios
+//
+//  Created by Nikita Timonin on 24/11/2019.
+//  Copyright © 2019 Tinode. All rights reserved.
+//
+
+import UIKit
+
+struct KeyboardInfo {
+    
+    let frameBegin: CGRect
+    let frameEnd: CGRect
+    let isLocal: Bool
+    let animationDuration: TimeInterval
+    let animationCurve: UIView.AnimationCurve
+    
+    init?(notification: Notification) {
+        guard
+            let userInfo = notification.userInfo,
+            let frameBegin = userInfo[UIResponder.keyboardFrameBeginUserInfoKey] as? CGRect,
+            let frameEnd = userInfo[UIResponder.keyboardFrameEndUserInfoKey] as? CGRect,
+            let isLocal = userInfo[UIResponder.keyboardIsLocalUserInfoKey] as? Bool,
+            let animationDuration = userInfo[UIResponder.keyboardAnimationDurationUserInfoKey] as? TimeInterval,
+            let rawCurve = userInfo[UIResponder.keyboardAnimationCurveUserInfoKey] as? Int,
+            let animationCurve = UIView.AnimationCurve(rawValue: rawCurve)
+            else { return nil }
+        
+        self.frameBegin = frameBegin
+        self.frameEnd = frameEnd
+        self.isLocal = isLocal
+        self.animationDuration = animationDuration
+        self.animationCurve = animationCurve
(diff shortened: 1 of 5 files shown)


Tinodios/MessageViewController+Keyboard.swift
```diff
@@ -10,49 +10,51 @@ import UIKit
...
+            object: nil)
     }
 
-    func removeKeyboardObservers() {
-        NotificationCenter.default.removeObserver(self, name: UIResponder.keyboardWillChangeFrameNotification, object: nil)
-        NotificationCenter.default.removeObserver(self, name: UITextView.keyboardWillHideNotification, object: nil)
-    }
-
-    @objc func adjustForKeyboard(notification: Notification) {
+    @objc func adjustForKeyboard(_ notification: Notification) {
```
aforge: From the UX perspective, it is desirable to preserve the content that the user views on the screen.
The invariant we attempt to maintain is preserving the bottom of the presented content, i.e. if the bottommost message visible to the user should remain the bottommost message regardless of the transformations. Hence, `contentBottom` var and its contribution to `contentOffset` in the original code.
Can you preserve this logic in this PR please?
Please, also ensure the correct behavior on device rotation.

Tinodios/KeyboardInfo.swift
```diff
@@ -0,0 +1,41 @@
+//
```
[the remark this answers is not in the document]
nikletsopen: Added in - https://github.com/tinode/ios/pull/109/commits/abea2f62265dc07193b6d917e226e42653a9e6c2

Tinodios/KeyboardInfo.swift
```diff
@@ -0,0 +1,41 @@
...
+//  Tinodios
+//
+//  Created by Nikita Timonin on 24/11/2019.
+//  Copyright © 2019 Tinode. All rights reserved.
+//
+
+import UIKit
+
+struct KeyboardInfo {
+
```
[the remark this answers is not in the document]
nikletsopen: Fixed in - https://github.com/tinode/ios/pull/109/commits/49ab36bfd37b41d62707aa2e2652747a2d5c2ab3

Tinodios/MessageViewController+Keyboard.swift
nikletsopen: In the original implementation collectionView scrolls all the way down to the bottom every time the keyboard is shown. It happens because of the `overlap` value is not calculated correctly in the case when `collectionView.contentSize.height > collectionView.bounds.height`. In all major messengers like WhatsApp, Telegram,etc  keyboard just lifts up the collectionView preserving its scroll position and does not go all the way down.  This PR is handling this case correctly.

nikletsopen: I am currently testing this solution in another project and it works well. Although I am also taking into consideration the height of `inputAccessoryView`, but you have it as a private var, so I've decided not to touch this part. Unfortunately I am having a problem with launching "Tinode" debug version on my device, so I can not really test it. On Sign Up I am getting the attached error. Could it be because of the incorrect signing or problems with "GoogleService-Info-Development.plist" file? 
photo_2019-11-25 11 44 38

or-else: >  inputAccessoryView, but you have it as a private var, so I've decided not to touch this part

You are welcome to change it.

or-else: > Could it be because of the incorrect signing or problems with "GoogleService-Info-Development.plist" file?

When running in a simulator it expects to connect to the server running at localhost:6060. Change the server address to sandbox.tinode.co.
https://github.com/tinode/ios/blob/master/Tinodios/Utils.swift#L249

aforge: > I am currently testing this solution in another project and it works well. Although I am also taking into consideration the height of `inputAccessoryView`, but you have it as a private var, so I've decided not to touch this part.

Yes, please do factor `inputAccessoryView` height into your calculations. Right now, if you pop and then close the keyboard, at the bottom of `MessageVC.collectionView` you will see unnecessary padding that's precisely `inputAccessoryView` height. It is a bug. So you can kill two birds with one stone by fixing it in this PR.

nikletsopen: added in - https://github.com/tinode/ios/pull/109/commits/b93c285f1e67cb846ed7847b0c9fb875f7d9206b

aforge: It looks like that leaves a certain amount of space at the bottom after you dismiss the keyboard. Can you take a look please?
Screen Shot 2019-11-26 at 9 29 55 PM

nikletsopen:
Possible fix in - https://github.com/tinode/ios/pull/109/commits/046f1b2f4e3a3000692941ab0a0841bdbed339f1

palatov:
@aforge  FYI, I've been facing some cases when this overall approach was not working as expected. So, I recommend to put it on hold, before I'll fix those faulty cases.

nikletsopen:
@aforge FYI, I've been facing some cases when this overall approach was not working as expected. So, I recommend to put it on hold, before I'll fix those faulty cases.

aforge:
@nikletsopen Putting on hold sounds good.
@Kuvonchbeky Thanks for testing.

Tinodios/MessageViewController+Keyboard.swift
nikletsopen: Fixed in - https://github.com/tinode/ios/pull/109/commits/dab989c3e2606397bd466685b7552c796c9dec99

nikletsopen:
> this will cause the message buble to go under the message send view (iOS 13.2.2)

Fixed in - https://github.com/tinode/ios/pull/109/commits/dab989c3e2606397bd466685b7552c796c9dec99

aforge:
Cool. Thanks for taking care of it.

PR #111, opened 2019-11-27
Add code to fetch messages in background on data notifications.

aforge:
This is a straw man version. More testing is yet to come.

### DIFF
--- a/TinodeSDK/Topic.swift
+++ b/TinodeSDK/Topic.swift
@@ -118,6 +118,11 @@ open class Topic<DP: Codable & Mergeable, DR: Codable & Mergeable, SP: Codable, 
     }
 
     public class MetaGetBuilder {
+
+        public enum Options {
+            case desc, sub, data(Int), tags, cred, del
+        }
+
         let topic: TopicProto
         var meta: MsgGetMeta
 
--- a/TinodeSDK/Topic.swift
+++ b/TinodeSDK/Topic.swift
@@ -118,11 +118,6 @@ open class Topic<DP: Codable & Mergeable, DR: Codable & Mergeable, SP: Codable, 
     }
 
     public class MetaGetBuilder {
-
-        public enum Options {
-            case desc, sub, data(Int), tags, cred, del
-        }
-
         let topic: TopicProto
         var meta: MsgGetMeta
 
--- a/Tinodios/AppDelegate.swift
+++ b/Tinodios/AppDelegate.swift
@@ -94,14 +94,24 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             if Utils.connectAndLoginSync() {
                 let tinode = Cache.getTinode()
                 var result: UIBackgroundFetchResult = .failed
-                if let topic = tinode.getTopic(topicName: topicName) as? DefaultComTopic {
-                    if (topic.seq ?? 0) >= seq {
-                        result = .noData
-                    } else if let msg = try? UiUtils.attachToTopic(topic: topic, fetchOptions: [.data(10), .del])?.getResult(), (msg.ctrl?.code ?? 500) < 300 {
-                        result = .newData
-                    }
+                defer { completionHandler(result) }
+                var topic: DefaultComTopic
+                if !tinode.isTopicTracked(topicName: topicName) {
+                    // New topic. Create it.
+                    guard let t = tinode.newTopic(for: topicName, with: nil) as? DefaultComTopic else { return }
+                    topic = t
+                    topic.persist(true)
+                } else {
+                    // Existing topic.
+                    guard let t = tinode.getTopic(topicName: topicName) as? DefaultComTopic else { return }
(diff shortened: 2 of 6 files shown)


Tinodios/AppDelegate.swift
```diff
@@ -92,22 +61,55 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+        guard let topicName = userInfo["topic"] as? String, !topicName.isEmpty else {
+            completionHandler(.failed)
+            return
+        }
+        if state == .background || (state == .inactive && !self.appIsStarting) {
+            if Utils.connectAndLoginSync() {
+                let tinode = Cache.getTinode()
+                var result: UIBackgroundFetchResult = .failed
+                if let topic = tinode.getTopic(topicName: topicName) as? DefaultComTopic {
+                    if let msg = try? UiUtils.attachToTopic(topic: topic, fetchTags: false, maxMessages: 10)?.getResult(), (msg.ctrl?.code ?? 500) < 300 {
```
or-else: The notification could be stale.  Maybe the client has already fetched the messages. I would compare `userInfo["seq"]` with the DB value before doing any network calls. Otherwise `attachToTopic` will send a useless `{sub}` request .

Tinodios/AppDelegate.swift
```diff
@@ -92,22 +61,55 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+                        result = .newData
+                    }
+                    tinode.logout()
```
or-else: Is it running in a different process than the main app? I.e. this is a different address space, right?

Tinodios/AppDelegate.swift
aforge: Yup, sounds reasonable.

Tinodios/AppDelegate.swift
aforge: Sure, you are right. I haven't tested this code since I need a server-side change first.

Tinodios/AppDelegate.swift
```diff
@@ -92,22 +61,56 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+        let state = application.applicationState
+        guard let topicName = userInfo["topic"] as? String, !topicName.isEmpty, let seq = userInfo["seq"] as? Int else {
+            completionHandler(.failed)
+            return
+        }
+        if state == .background || (state == .inactive && !self.appIsStarting) {
+            if Utils.connectAndLoginSync() {
+                let tinode = Cache.getTinode()
+                var result: UIBackgroundFetchResult = .failed
+                if let topic = tinode.getTopic(topicName: topicName) as? DefaultComTopic {
```
or-else: This could be a message from an entirely new topic. It might make sense to handle it too.

Tinodios/UiUtils.swift
```diff
@@ -129,7 +129,20 @@ class UiUtils {
             fnd.subscribe(set: nil, get: nil) :
             PromisedReply<ServerMessage>(value: ServerMessage())
     }
-
+    public static func attachToTopic(topic: DefaultComTopic?, fetchTags: Bool, maxMessages: Int) -> PromisedReply<ServerMessage>? {
+        guard let topic = topic else { return nil }
+        var builder = topic.getMetaGetBuilder()
+            .withDesc()
```
or-else: You are fetching description which you don't need.

Tinodios/UiUtils.swift
```diff
@@ -129,7 +129,20 @@ class UiUtils {
...
+            .withSub()
```
or-else: Ditto. You don't need subscription.

Tinodios/AppDelegate.swift
aforge: Added code to handle it.

Tinodios/UiUtils.swift
aforge: Made this optional.

Tinodios/UiUtils.swift
aforge: Done.

TinodeSDK/Topic.swift
```diff
@@ -118,6 +118,11 @@ open class Topic<DP: Codable & Mergeable, DR: Codable & Mergeable, SP: Codable,
     }
 
     public class MetaGetBuilder {
+
+        public enum Options {
+            case desc, sub, data(Int), tags, cred, del
```
or-else: Why not just build what you want and pass it to the call? I.e. instead of 
```
Utils.attachToTopic(topic: topic, fetchOptions: [.data(10), .del])
```
use
```
Utils.attachToTopic(topic: topic, fetch: topic.getMetaGetBuilder().withData(...).withDel())
```
Enum trades flexibility of builder for a little bit of brevity but less clarity.

aforge: Done.

PR #115, opened 2019-11-29
Handle universal links (to *.tinode.co).

aforge:
This change will also need configuration on the server.

### DIFF
--- a/Tinodios/AppDelegate.swift
+++ b/Tinodios/AppDelegate.swift
@@ -126,6 +126,21 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             completionHandler(.noData)
         }
     }
+
+    func application(_ application: UIApplication,
+                     continue userActivity: NSUserActivity,
+                     restorationHandler: @escaping ([UIUserActivityRestoring]?) -> Void) -> Bool {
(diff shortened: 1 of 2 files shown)


Tinodios/AppDelegate.swift
```diff
@@ -126,6 +126,21 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+
+    func application(_ application: UIApplication,
+                     continue userActivity: NSUserActivity,
+                     restorationHandler: @escaping ([UIUserActivityRestoring]?) -> Void) -> Bool {
+        guard userActivity.activityType == NSUserActivityTypeBrowsingWeb,
+            let url = userActivity.webpageURL,
+            let components = URLComponents(url: url, resolvingAgainstBaseURL: true) else {
+            return false
+        }
+        if (components.host?.hasSuffix("tinode.co") ?? false) {
```
or-else: OK for now but we need a better way to handle it in the future: 
1. Support for third party URLs.
2. Differentiate between opening prod/sandbox URLs.

aforge: Added a todo.

PR #116, opened 2019-12-01
Display licenses in Settings bundle.

aforge:
Also, update pods.

Licences are stored in Settings.bundle/Acknowledgements.plist which is can be generated by running `pod update`.


Scripts/set_default_preferences.sh
```diff
@@ -12,39 +12,47 @@ info_plist="$TARGET_BUILD_DIR/$INFOPLIST_PATH"
...
 fi
 # Find key for TLS use
 find_array_entry "use_tls_preference"
 tls_index=$?
-if [ "$tls_index" -eq 0 ]; then
+if [ "$tls_index" -eq -1 ]; then
   echo "Entry use_tls_preference not found"
   exit 1
 fi
+# Find key for TLS use
```
or-else: Wrong comment.

PR #121, opened 2019-12-10
Fix background data fetch issues.

aforge:
Data is now fetched when the app is in background and foreground.
Fixed a problem in Topic subscription handling which rewrote Topic.description.acs.

### DIFF
--- a/Tinodios/AppDelegate.swift
+++ b/Tinodios/AppDelegate.swift
@@ -81,44 +81,58 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         application.applicationIconBadgeNumber = Cache.totalUnreadCount()
     }
 
+    // Synchronously connects to topic |topicName| and fetches its messages
+    // if the last received message was prior to |seq|.
+    @discardableResult
+    private func fetchData(for topicName: String, seq: Int) -> UIBackgroundFetchResult {
+        let tinode = Cache.getTinode()
+        var topic: DefaultComTopic
+        var builder: DefaultComTopic.MetaGetBuilder
+        if !tinode.isTopicTracked(topicName: topicName) {
+            // New topic. Create it.
+            guard let t = tinode.newTopic(for: topicName, with: nil) as? DefaultComTopic else { return .failed }
+            topic = t
+            topic.persist(true)
+            builder = topic.getMetaGetBuilder().withDesc().withSub()
+        } else {
+            // Existing topic.
+            guard let t = tinode.getTopic(topicName: topicName) as? DefaultComTopic else { return .failed }
+            topic = t
+            builder = topic.getMetaGetBuilder()
+        }
+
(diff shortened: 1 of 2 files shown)


Tinodios/AppDelegate.swift
```diff
@@ -81,44 +81,58 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+    @discardableResult
+    private func fetchData(for topicName: String, seq: Int) -> UIBackgroundFetchResult {
+        let tinode = Cache.getTinode()
+        var topic: DefaultComTopic
+        var builder: DefaultComTopic.MetaGetBuilder
+        if !tinode.isTopicTracked(topicName: topicName) {
+            // New topic. Create it.
+            guard let t = tinode.newTopic(for: topicName, with: nil) as? DefaultComTopic else { return .failed }
+            topic = t
+            topic.persist(true)
```
or-else: This is different from Android. Maybe it make sense to do it the same way here too: it seems more logical to handle topic persistence in the SDK. 

I don't suggest to make the change in this PR. It should be a separate PR.

Tinodios/AppDelegate.swift
```diff
@@ -81,44 +81,58 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+            // Existing topic.
+            guard let t = tinode.getTopic(topicName: topicName) as? DefaultComTopic else { return .failed }
+            topic = t
+            builder = topic.getMetaGetBuilder()
+        }
+
+        if (topic.recv ?? 0) >= seq {
+            return .noData
+        }
+        defer {
```
or-else: What of topic is already attached?

Tinodios/AppDelegate.swift
```diff
@@ -81,44 +81,58 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+            if topic.attached {
+                topic.leave()
+            }
+        }
+        if let msg = try? topic.subscribe(
```
or-else: This will produce an unnecessary network traffic and will fail if topic is already attached.

Tinodios/AppDelegate.swift
aforge: Yes, will make a follow-up PR.

Tinodios/AppDelegate.swift
aforge: This function is called when (a) the app is in the foreground but the app isn't attached to the current topic (see func userNotificationCenter(_,willPresent,withCompletionHandler)) and (b) the app is in the background (no topic is attached) or not running (see func application(application:didReceiveRemoteNotification:fetchCompletionHandler)).

Tinodios/AppDelegate.swift
aforge: See above.

PR #122, opened 2019-12-11
A couple more fixes

aforge:
1. Move topic persistence logic to the SDK when we create a new topic (from AppDelegate and FindVC).
2. Added FB analytics package (now the app stats are available in the FB dashboard).
3. Prevent sendMessageBar from disappearing when the user dismisses image picker.


Podfile
```diff
@@ -22,6 +22,7 @@ end
 def ui_pods
     pod 'SwiftKeychainWrapper', '~> 3.2'
     pod 'Firebase/Messaging'
+    pod 'Firebase/Analytics'
```
or-else: We should update the docs with respect to the list of dependencies: mention FB + PhoneNumberKit

TinodeSDK/Tinode.swift
```diff
@@ -616,6 +616,12 @@ public class Tinode {
     public func newTopic(for name: String, with listener: DefaultTopic.Listener?) -> TopicProto {
         return Tinode.newTopic(withTinode: self, forTopic: name, withListener: listener)
     }
+    public func createAndPresistTopic(withName name: String,
```
or-else: It seems like topic is automatically persisted when created with non-nil Tinode reference:
https://github.com/tinode/ios/blob/master/TinodeSDK/Topic.swift#L412

This is exactly the same way as in Tindroid:
https://github.com/tinode/tindroid/blob/master/tinodesdk/src/main/java/co/tinode/tinodesdk/Topic.java#L88

Why is it necessary to explicitly persist it in iOS and not in Android?

aforge: `Tinode.startTackingTopic()` only adds it to in-memory data structures.
`Topic.persist()` writes it to sqlitedb.
The two should not always be called together. E.g. when you load topics from sqlitedb you only need the former but not the latter.
https://github.com/tinode/tindroid/blob/master/app/src/main/java/co/tinode/tindroid/db/TopicDb.java#L383

Podfile
aforge: Yes, it is done.

TinodeSDK/Tinode.swift
or-else: Actually, I was wrong. Android also just adds it to in-memory structures in https://github.com/tinode/tindroid/blob/master/tinodesdk/src/main/java/co/tinode/tinodesdk/Topic.java#L88

It persists it on first subscription:
https://github.com/tinode/tindroid/blob/master/tinodesdk/src/main/java/co/tinode/tinodesdk/Topic.java#L712

or-else: I dislike `createAndPresistTopic` because it exposes cache management to the client.

README.md
```diff
@@ -48,7 +48,8 @@ The immediate goal is to have a basic end-to-end working application prototype t
 
 * https://github.com/stephencelis/SQLite.swift for local persistence
 * https://github.com/jrendel/SwiftKeychainWrapper
-* [GoogleService-Info.plist](https://firebase.google.com/docs/cloud-messaging/ios/client) for push notifications. See below.
+* https://github.com/marmelroy/PhoneNumberKit
+* GoogleService-Info.plist for [push notifications](https://firebase.google.com/docs/cloud-messaging/ios/client) and [analytics](https://firebase.google.com/docs/analytics/get-started?platform=ios). See below.
```
or-else: `GoogleService-Info.plist` -> `Google Firebase`?

TinodeSDK/Tinode.swift
aforge: Removed `createAndPresistTopic`.

README.md
aforge: Done.

PR #123, opened 2019-12-12
added <head> column to the database

Kuvonchbeky:


or-else:
Please increment database version

https://github.com/tinode/ios/blob/devel/TinodiosDB/BaseDb.swift#L14

Kuvonchbeky:
Database version has been incremented

TinodiosDB/SqlStore.swift
```diff
@@ -195,6 +195,8 @@ public class SqlStore : Storage {
         msg.seq = 0
         msg.status = initialStatus
         msg.content = data
+        //FIXME: probably in the future this should change for reply/forwarded messages
+        msg.head = data.isPlain ? nil : ["mime": JSONValue.string(Drafty.kMimeType)]
```
or-else: I don't think this is necessary.

Kuvonchbeky: Unless add this then <head> doesn't save on the local database

or-else: Then there is bug somewhere. This line just masks the problem, it does not solve it.

or-else: Let's wait for @aforge. He knows this code better than me.

Kuvonchbeky: For now this same this line: https://github.com/tinode/ios/blob/master/TinodeSDK/Tinode.swift#L1071
How can the problem be solved?

or-else: > For now this same this line:

That's different. That's the message being sent to the server. You need to create it before sending, including setting the header.

What you are saving to DB is a message received from the server. It's already fully formed. You should not need to modify the header.

or-else: As for how to solve it, are you sure there is a problem? It should work without forcing the header.

Kuvonchbeky: > As for how to solve it, are you sure there is a problem?

I'm not sure there a problem. You said "This line just masks the problem, it does not solve it."

> Let's wait for @aforge. He knows this code better than me.

Okay

or-else: > I'm not sure there a problem

You said: _Unless add this then "head" hasn't save on the local database_. So, I assume you think there is a problem and I asked you if you are sure. Did you receive a message with non-nil `head` from the server and when it was saved the `head` was `null` in the database?

Kuvonchbeky: I'm sure the `head` doesn't save on the local database of the sender if I don't add that line. But on the local database of the receiver the `head` has saved.

or-else: OK, now I understand: the incoming messages get saved correctly, but the outgoing messages are saved without the `head`. That's a bug. It should be fixed, for instance, here 
https://github.com/tinode/ios/blob/master/TinodeSDK/Topic.swift#L1103 
and here
https://github.com/tinode/ios/blob/master/TinodeSDK/model/ClientMessages.swift#L395

like it's done in Android:
https://github.com/tinode/tindroid/blob/master/tinodesdk/src/main/java/co/tinode/tinodesdk/model/MsgClientPub.java#L26

Let's wait for aforge.

Your code is incorrect, please remove the line.

Kuvonchbeky: > like it's done in Android:

Android has the same problem. I'm sure

aforge: Agree with @or-else . This is not a good place for message modification.
Can you modify https://github.com/tinode/ios/blob/master/TinodeSDK/model/ClientMessages.swift#L395 please?

aforge: @Kuvonchbeky, please hold off from making changes. I'm looking into it now.

Kuvonchbeky: > Can you modify https://github.com/tinode/ios/blob/master/TinodeSDK/model/ClientMessages.swift#L395 please?

I think this don't affect to storing "`head`" on the localDB of the sender. It will anyway equals `nil`. Because, as @or-else said it's modified in Android, but it didn't solve the problem

or-else: @aforge will fix the other problem. Just remove line2 198 and 199 from your PR and I will approve.

or-else:
JFYI, we are making the changes to save `head`. First Android https://github.com/tinode/tindroid/pull/74, then iOS, then we will merge your PR.

Kuvonchbeky:
> JFYI, we are making the changes to save `head`. First Android [tinode/tindroid#74](https://github.com/tinode/tindroid/pull/74), then iOS, then we will merge your PR.

I see, thanks.

PR #124, opened 2019-12-14
Pass message headers to Tinode.publish explicitly.

aforge:
Needed in order to correctly save message headers for outgoing messages.

### DIFF
--- a/TinodeSDK/Tinode.swift
+++ b/TinodeSDK/Tinode.swift
@@ -1060,8 +1060,9 @@ public class Tinode {
         return sendWithPromise(payload: msg, with: msgId)
     }
 
-    public static func draftyHeaders(for message: Drafty) -> [String: JSONValue] {
-        var head = [String: JSONValue]()
+    public static func draftyHeaders(for message: Drafty) -> [String: JSONValue]? {
+        guard !message.isPlain else { return nil }
+        var head: [String: JSONValue] = [:]
         head["mime"] = JSONValue.string(Drafty.kMimeType)
         if let refs = message.getEntReferences() {
             head["attachments"] = JSONValue.array(refs.map { JSONValue.string($0) })
--- a/TinodeSDK/Tinode.swift
+++ b/TinodeSDK/Tinode.swift
@@ -1060,18 +1060,28 @@ public class Tinode {
         return sendWithPromise(payload: msg, with: msgId)
     }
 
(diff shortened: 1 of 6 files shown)


TinodeSDK/Tinode.swift
```diff
@@ -1060,18 +1060,28 @@ public class Tinode {
         return sendWithPromise(payload: msg, with: msgId)
     }
 
-    internal func publish(topic: String, head: [String:JSONValue]?, content: Drafty) -> PromisedReply<ServerMessage>? {
+    public static func draftyHeaders(for message: Drafty) -> [String: JSONValue] {
+        var head = [String: JSONValue]()
```
or-else: Swift 4 and 5 canonical is 
`var head: [String:JSONValue] = []`

TinodeSDK/Tinode.swift
```diff
@@ -1060,18 +1060,28 @@ public class Tinode {
         return sendWithPromise(payload: msg, with: msgId)
     }
 
-    internal func publish(topic: String, head: [String:JSONValue]?, content: Drafty) -> PromisedReply<ServerMessage>? {
+    public static func draftyHeaders(for message: Drafty) -> [String: JSONValue] {
```
or-else: Just a question. Should we check `message.isPlain` and skip creating headers if true?

TinodeSDK/Tinode.swift
```diff
@@ -1060,18 +1060,28 @@ public class Tinode {
...
+    public func publish(topic: String, head: [String:JSONValue]?, content: Drafty) -> PromisedReply<ServerMessage>? {
         let msgId = getNextMsgId()
         let msg = ClientMessage<Int, Int>(
             pub: MsgClientPub(id: msgId, topic: topic, noecho: true, head: head, content: content))
         return sendWithPromise(payload: msg, with: msgId)
     }
-
+    /*
     public func publish(topic: String, data: Drafty) -> PromisedReply<ServerMessage>? {
         return publish(topic: topic,
```
or-else: Delete?

TinodeSDK/Tinode.swift
aforge: Sure. Originally, that's how the code was written.
I made it so in order to be consistent with Android.

TinodeSDK/Tinode.swift

TinodeSDK/Tinode.swift
aforge: This should be gone. A stale snapshot.

PR #126, opened 2019-12-22
Make upload progress bar to circular.

aforge:
Similar to Whatsapp uploads.

### DIFF
--- a/Tinodios/MessageCell.swift
+++ b/Tinodios/MessageCell.swift
@@ -107,10 +107,7 @@ class MessageCell: UICollectionViewCell {
         return label
     }()
 
-    var progressView: CircularProgressView = {
-        let progress = CircularProgressView()
-        return progress
-    }()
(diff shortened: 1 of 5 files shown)


Tinodios/MessageCell.swift
```diff
@@ -107,18 +107,9 @@ class MessageCell: UICollectionViewCell {
...
-    }()
-
-    var cancelUploadButton: UIButton = {
-        let button = UIButton()
-        button.setTitle("×", for: .normal)
-        button.setTitleColor(UIColor.blue, for: .normal)
-        button.titleLabel?.font = UIFont.systemFont(ofSize: 10)
-        return button
+    var progressView: CircularProgressView = {
+        let progress = CircularProgressView()
```
or-else: Unnecessary assignment?

aforge: Fixed.

PR #128, opened 2019-12-25
Take care of crash & keyboard layout issue

aforge:


TinodeSDK/Tinode.swift
```diff
@@ -993,6 +992,7 @@ public class Tinode {
         guard let endpointURL = self.channelsURL(useWebsocketProtocol: true) else {
             throw TinodeError.invalidState("Could not form server url.")
         }
+        nextMsgId = 0xffff + Int((Float(arc4random()) / Float(UInt32.max)) * 0xffff)
```
or-else: resetMsgId() or initMsgId()?

PR #130, opened 2019-12-29
Message deletion logic in Tinode SDK and DB packages

aforge:
Modifications similar to github.com/tinode/tindroid/pull/75.
UI changes to follow.

### DIFF
--- a/TinodeSDK/model/MsgRange.swift
+++ b/TinodeSDK/model/MsgRange.swift
@@ -10,20 +10,20 @@ import Foundation
 // Represents a contiguous range of message sequence ids:
 // inclusive on the left - exclusive on the right.
 public class MsgRange: Codable {
-    public var low: Int?
+    public var low: Int
     public var hi: Int?
-    public var lower: Int { return low ?? 0 }
+    public var lower: Int { return low }
     public var upper: Int { return hi ?? lower + 1 }
 
     public init() {
         self.low = 0
         self.hi = nil
     }
-    public init(id: Int?) {
+    public init(id: Int) {
         self.low = id
         self.hi = nil
     }
-    public init(low: Int?, hi: Int?) {
+    public init(low: Int, hi: Int?) {
         self.low = low
         self.hi = hi
     }
@@ -37,7 +37,7 @@ public class MsgRange: Codable {
     }
     public func compare(to other: MsgRange) -> Int {
         let rl = MsgRange.cmp(self.low, other.low)
-        return rl == 0 ? MsgRange.cmp(self.hi, other.hi) : rl
+        return rl == 0 ? MsgRange.cmp(other.hi, self.hi) : rl
     }
     /**
      * Collapse multiple possibly overlapping ranges into as few ranges non-overlapping
--- a/TinodeSDK/model/MsgRange.swift
+++ b/TinodeSDK/model/MsgRange.swift
@@ -1,0 +1,94 @@
+//
+//  MsgRange.swift
+//  TinodeSDK
+//
+//  Copyright © 2019 Tinode. All rights reserved.
+//
+
+import Foundation
+
+// Represents a contiguous range of message sequence ids:
+// inclusive on the left - exclusive on the right.
+public class MsgRange: Codable {
+    public var low: Int?
+    public var hi: Int?
+    public var lower: Int { return low ?? 0 }
+    public var upper: Int { return hi ?? lower + 1 }
+
+    public init() {
+        self.low = 0
+        self.hi = nil
+    }
+    public init(id: Int?) {
+        self.low = id
+        self.hi = nil
+    }
+    public init(low: Int?, hi: Int?) {
+        self.low = low
+        self.hi = hi
+    }
+    public init(from another: MsgRange) {
+        self.low = another.low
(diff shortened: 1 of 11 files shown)


TinodeSDK/model/MsgRange.swift
```diff
@@ -0,0 +1,94 @@
...
+        self.low = another.low
+        self.hi = another.hi
+    }
+
+    private static func cmp(_ val1: Int?, _ val2: Int?) -> Int {
+        return (val1 ?? 0) - (val2 ?? 0)
+    }
+    public func compare(to other: MsgRange) -> Int {
+        let rl = MsgRange.cmp(self.low, other.low)
+        return rl == 0 ? MsgRange.cmp(self.hi, other.hi) : rl
```
or-else: This should be the other way around `MsgRange.cmp(other.hi, self.hi)`: sort order is longer ranges come first.

TinodiosDB/SqlStore.swift
```diff
@@ -255,30 +255,45 @@ public class SqlStore : Storage {
...
         guard let st = topic.payload as? StoredTopic, let topicId = st.id else { return false }
+        let idHi = idHi <= 0 ? (st.maxLocalSeq ?? 0) + 1 : idHi
+        var success = false
         do {
-            var result = false
-            try dbh?.db?.savepoint("SqlStore.msgDelete") {
-                result = (dbh?.topicDb?.msgDeleted(topic: topic, delId: id) ?? false) &&
-                    (dbh?.messageDb?.delete(topicId: topicId, from: idLo, to: idHi) ?? false)
+            try dbh?.db?.savepoint("SqlStore.msgDelete-bounds") {
+                success = (delId >= 0 || (dbh?.topicDb?.msgDeleted(topic: topic, delId: delId, from: idLo, to: idHi) ?? false)) &&
```
or-else: This is exactly the same as in Android but I think `delId >= 0` is a bug in Android in the first place. It should be `delId <= 0`. And maybe even that is unnecessary: `msgDelete` is called 3 times only from https://github.com/tinode/ios/blob/0014991e8bd0b54c73a6fb178ed5c855a9489163/TinodeSDK/Topic.swift and in all cases the caller ensures `delId > 0`.

The logic here is to update denormalized min/max seq values in Topic only when the update is synchronized with the server. Which means delId should be non-zero for `dbh?.topicDb?.msgDeleted` to be called.

I'll update and test Android right after I finish this review.

TinodiosDB/SqlStore.swift
```diff
@@ -255,30 +255,45 @@ public class SqlStore : Storage {
...
-        return false
+    public func msgDelete(topic: TopicProto, delete delId: Int, deleteAllIn ranges: [MsgRange]?) -> Bool {
+        guard let st = topic.payload as? StoredTopic, let topicId = st.id,
+            let ranges = ranges, !ranges.isEmpty else { return false }
+        let collapsedRanges = MsgRange.collapse(ranges)
+        let enclosing = MsgRange.enclosing(for: collapsedRanges)
+        var success = false
+        do {
+            try dbh?.db?.savepoint("SqlStore.msgDelete-ranges") {
+                success = (delId >= 0 || (dbh?.topicDb?.msgDeleted(topic: topic, delId: delId, from: enclosing!.lower, to: enclosing!.upper) ?? false)) &&
```
or-else: It has the same `delId >= 0` issue.

TinodiosDB/SqlStore.swift
or-else: I removed the check on Android:
https://github.com/tinode/tindroid/commit/3d914569b80dfa94a64a9090a6d2de8ec572c1ee
It seems to work correctly.

or-else: I added this check in one of the earlier iterations and apparently forgot to remove it.

TinodeSDK/model/MsgRange.swift
```diff
@@ -0,0 +1,94 @@
...
+//
+//  Copyright © 2019 Tinode. All rights reserved.
+//
+
+import Foundation
+
+// Represents a contiguous range of message sequence ids:
+// inclusive on the left - exclusive on the right.
+public class MsgRange: Codable {
+    public var low: Int?
```
or-else: It just crossed my mind that it can be just `Int` as opposite to `Int?`. I can also safely make it `int` on Android (as opposite to `Integer`).

TinodeSDK/model/MsgRange.swift
aforge: Good catch!

TinodiosDB/SqlStore.swift
aforge: Ah, okay. I'll remove it too.

TinodiosDB/SqlStore.swift
aforge: Done.

TinodeSDK/model/MsgRange.swift
aforge: Yes, totally. I was playing with the function signatures where `low` is present but decided to keep it `Int?` for consistency reasons. Let's change it to `Int`.

PR #131, opened 2020-01-01
Display deleted messages in MessageVC.

aforge:


Tinodios/MessageBubbleDecorator.swift
```diff
@@ -102,6 +102,30 @@ class MessageBubbleDecorator {
         }
     }
 
+    public static func drawDeleted(_ rect: CGRect) -> UIBezierPath {
```
or-else: Please use `UIBezierPath(roundedRect:,cornerRadius:)` instead:
```
return UIBezierPath(roundedRect: CGRect(x: 0, y: 0, width: size.width, height: size.height), cornerRadius: 7.61)
```

Tinodios/MessageViewController.swift
```diff
@@ -635,9 +636,10 @@ extension MessageViewController: UICollectionViewDataSource {
     private func configureCell(cell: MessageCell, with message: Message, at indexPath: IndexPath) {
 
         cell.seqId = message.seqId
+        cell.isDeleted = message.isDeleted
 
         cell.content.backgroundColor = nil
-        if isFromCurrentSender(message: message) {
+        if message.isDeleted || isFromCurrentSender(message: message) {
```
or-else: Make the bubble color different, like in Android and web?

Tinodios/MessageBubbleDecorator.swift
aforge: Nice!

Tinodios/MessageViewController.swift

PR #132, opened 2020-01-03
Logic to detect sequence id gaps in cached messages.

aforge:
Similar to https://github.com/tinode/tindroid/pull/77.
Starting this early.

### DIFF
--- a/TinodeSDK/Topic.swift
+++ b/TinodeSDK/Topic.swift
@@ -133,8 +133,7 @@ open class Topic<DP: Codable & Mergeable, DR: Codable & Mergeable, SP: Codable, 
         }
         public func withEarlierData(limit: Int?) -> MetaGetBuilder {
             if let r = topic.missingMessageRange, r.low >= 1 {
-                let totalMissing = r.upper - r.lower
-                return withData(since: nil, before: r.upper, limit: min(totalMissing, limit ?? Int.max))
+                return withData(since: r.lower, before: r.upper, limit: limit)
             }
             return withData(since: nil, before: nil, limit: limit)
         }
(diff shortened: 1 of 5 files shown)


TinodeSDK/Topic.swift
```diff
@@ -131,8 +132,9 @@ open class Topic<DP: Codable & Mergeable, DR: Codable & Mergeable, SP: Codable,
             return self
         }
         public func withEarlierData(limit: Int?) -> MetaGetBuilder {
-            if let r = topic.cachedMessageRange {
-                return withData(since: nil, before: r.low != 0 ? r.low : nil, limit: limit)
+            if let r = topic.missingMessageRange, r.low >= 1 {
+                let totalMissing = r.upper - r.lower
+                return withData(since: nil, before: r.upper, limit: min(totalMissing, limit ?? Int.max))
```
or-else: I did it differently on Android:
```
withData(since: r.lower, before: r.upper, limit: limit)
```
Server will do the right thing when the request parameters are ambigous.

TinodeSDK/Topic.swift
```diff
@@ -366,9 +368,11 @@ open class Topic<DP: Codable & Mergeable, DR: Codable & Mergeable, SP: Codable,
...
 
     public var cachedMessageRange: MsgRange? {
-        get {
-            return store?.getCachedMessagesRange(topic: self)
-        }
+        return store?.getCachedMessagesRange(topic: self)
+    }
+
+    public var missingMessageRange: MsgRange? {
+        return (seq ?? 0) > 0 ? store?.getNextMissingRange(topic: self) : cachedMessageRange
```
or-else: This is another mistake I made in Android. `cachedMessageRange` returns *present* messages. So, instead of returning `cachedMessageRange` it should be those outside of what's returned by `cachedMessageRange`.

or-else: I think a simple 
```
return store?.getNextMissingRange(topic: self)
```
should be enough here.

aforge: Done.

TinodeSDK/Topic.swift
aforge: Done.

PR #136, opened 2020-01-20
Document previews, minor fixes in image previews.

or-else:


Tinodios/ImagePreviewController.swift
```diff
@@ -161,7 +161,7 @@ extension ImagePreviewController : SendImageBarDelegate {
         }
 
         // This notification is received by the MessageViewController.
-        NotificationCenter.default.post(name: Notification.Name("SendImageMessage"), object: msg)
+        NotificationCenter.default.post(name: Notification.Name("SendDraftyMessage"), object: msg)
```
aforge: Can you define notification names as constants in their respective view controllers please?

or-else: done

PR #137, opened 2020-01-23
Password visibility switches

aforge:


Tinodios/Base.lproj/Main.storyboard
```diff
@@ -2274,6 +2346,8 @@
         <image name="delete-message-22" width="22.5" height="22.5"/>
         <image name="edit-48" width="48" height="48"/>
         <image name="email-25" width="25" height="25"/>
+        <image name="eye.fill" catalog="system" width="64" height="38"/>
```
or-else: warning: SF Symbol 'eye.fill' is unavailable prior to iOS 13.0. [5]

Tinodios/Base.lproj/Main.storyboard
```diff
@@ -2274,6 +2346,8 @@
...
+        <image name="eye.slash.fill" catalog="system" width="64" height="40"/>
```
or-else: warning: SF Symbol 'eye.slash.fill' is unavailable prior to iOS 13.0. [5]

Tinodios/Base.lproj/Main.storyboard
aforge: Thanks for pointing out. Substituted with custom icons.

aforge:
Changes made as requested.

aforge:
Sounds reasonable. A good yak shave :-)

PR #138, opened 2020-01-24
Send explicit {note what=read} upon message receipt.

aforge:


Tinodios/MessageInteractor.swift
```diff
@@ -392,6 +392,11 @@ class MessageInteractor: DefaultComTopic.Listener, MessageBusinessLogic, Message
     }
     override func onData(data: MsgServerData?) {
         self.loadMessages()
+        if let from = data?.from, let seq = data?.seq, !Cache.getTinode().isMe(uid: from) {
+            DispatchQueue.main.asyncAfter(deadline: .now() + .seconds(1)) {
+                self.sendReadNotification(explicitSeq: seq)
```
or-else: I suspect the events placed into the main dispatch queue survive lifecycle events, like VC being removed. That means the read notification is going to be sent even when the VC is closed immediately. Or it's going to fire even if topic changes. 1 second is a short window, it's unlikely to happen, but still.

Tinodios/MessageInteractor.swift
```diff
@@ -392,6 +392,11 @@ class MessageInteractor: DefaultComTopic.Listener, MessageBusinessLogic, Message
...
+            }
```
or-else: Another problem with this approach (and it's also a problem on Android) is that if the user receives a large batch of messages, each one of them is going to be acknowledged separately even if only one ack is needed.

or-else: I fixed it like this: https://github.com/tinode/tindroid/commit/c5fe504c4b0ecde5e4c3061be331f35726e3edb7

aforge: Modified iOS correspondingly.

Tinodios/MessageInteractor.swift
aforge: Make reference to `self` weak.

Tinodios/MessageViewController.swift
```diff
@@ -336,33 +334,14 @@ class MessageViewController: UIViewController {
...
-                    self.processNotifications()
-                })
-        } else {
-            // Fallback on earlier versions
-            self.noteTimer = Timer.scheduledTimer(
-                timeInterval: 1,
-                target: self,
-                selector: #selector(self.processNotifications), userInfo: nil, repeats: true)
-        }
+        self.interactor?.sendReadNotification(explicitSeq: nil, when: .now())
```
or-else: think it should be `when: .now() + .seconds(1)`

Tinodios/MessageInteractor.swift
```diff
@@ -193,8 +196,18 @@ class MessageInteractor: DefaultComTopic.Listener, MessageBusinessLogic, Message
         }
         return true
     }
-    func sendReadNotification() {
-        topic?.noteRead()
+    func sendReadNotification(explicitSeq: Int? = nil, when deadline: DispatchTime = .now()) {
```
or-else: It looks like `sendReadNotification` is never called in a way which makes use of ` = .now()`.

aforge: Removed.

Tinodios/MessageViewController.swift

PR #143, opened 2020-03-08
Fix crash in Topic.loadSubs. Switch to debug builds by default.

aforge:


TinodeSDK/Topic.swift
```diff
@@ -567,7 +567,11 @@ open class Topic<DP: Codable & Mergeable, DR: Codable & Mergeable, SP: Codable,
         subsLastUpdated = loaded.max(by: {(s1, s2) -> Bool in
             ((s1.updated ?? Date.distantPast) < (s2.updated ?? Date.distantPast))
         })?.updated
-        subs = (Dictionary(uniqueKeysWithValues: loaded.map { ($0.user, $0) }) as! [String : Subscription<SP, SR>])
+        let filtered = loaded.filter { $0.user != nil }
+        if filtered.count != loaded.count {
+            Tinode.log.error("Some subscriptions missing user field (topic %@)", self.name)
+        }
+        subs = (Dictionary(uniqueKeysWithValues: filtered.map { ($0.user, $0) }) as! [String : Subscription<SP, SR>])
```
or-else: Would it be more logical to add a default value here than to filter out nil users? 
It would also be nice to find the place where invalid user is being inserted. It does not happen on Android.

Tinodios.xcodeproj/project.pbxproj
```diff
@@ -971,7 +971,7 @@
 				0A0380E521B3CC8200A3FA0E /* Release */,
 			);
 			defaultConfigurationIsVisible = 0;
-			defaultConfigurationName = Release;
+			defaultConfigurationName = Debug;
```
or-else: I'm not sure this is needed. The other crash seems to be the same as this one. The stack trace although showing (missing) looks otherwise the same.

aforge: This is for the ease of development mainly.
The release configuration points at api.tinode.co.

TinodeSDK/Topic.swift
aforge: Removed this code in favor of foreign key checks.

TinodiosDB/BaseDb.swift
```diff
@@ -93,6 +93,7 @@ public class BaseDb {
         self.topicDb!.createTable()
         self.subscriberDb!.createTable()
         self.messageDb!.createTable()
+        try! self.db!.run("PRAGMA foreign_key_check")
```
or-else: This call just checks for existing FK violations, but it does not set FK enforcement:
https://www.sqlite.org/pragma.html#pragma_foreign_key_check

_The foreign_key_check pragma checks the database, or the table called "table-name", for foreign key constraints that are violated and returns one row of output for each violation._

I think this is needed:
https://www.sqlite.org/pragma.html#pragma_foreign_keys
```
PRAGMA foreign_keys = boolean;
```

aforge: Good catch! Done.

PR #144, opened 2020-04-09
Incognito mode

or-else:
1. Added incognito mode (muted 'me' topic) to keep iOS app on par with Android and web
2. Split MeTopic, FndTopic, and CommTopic into separate files.

While doing this I discovered that:
1. The credential management is incomplete. It's possible to add new phone or email, but it's impossible to edit or delete and even view existing emails/phones.
https://github.com/tinode/ios/blob/master/Tinodios/AccountSettingsViewController.swift#L183 -- no attempt to pre-fill the field with existing data.
2. There are lots of places with unnecessary error checks like this:  
https://github.com/tinode/ios/blob/master/Tinodios/AccountSettingsViewController.swift#L204
`setMeta` does not throw:
https://github.com/tinode/ios/blob/master/TinodeSDK/Topic.swift#L874
and even this `do` is probably unnecessary:
https://github.com/tinode/ios/blob/master/TinodeSDK/Topic.swift#L875

I can fix credentials first. I can clean up number 2 as I go.


aforge:
I'm not sure if it's possible to get rid of checks easily: the promise reply handler is marked `throws` and the compiler requires a try-catch block.

or-else:
> I'm not sure if it's possible to get rid of checks easily

We could crash instead of throwing. Android crashes: `throw IllegalStateException` crashes the current thread. It's a `RuntimeException` which is unchecked:
https://docs.oracle.com/javase/7/docs/api/java/lang/RuntimeException.html

iOS does not have an equivalent but offers  `fatalError()`.

PR #145, opened 2020-04-11
Cleanup of PromisedReplies, credentials work now

or-else:
* Got rid of do ... try ... catch everywhere where PromisedReply is returned.
* Got rid of all optional returns with PromisedReply: `-> PromisedReply<ServerMessage>?` replaced with `-> PromisedReply<ServerMessage>`.
* Got rid of unnecessary arg names in `PromisedReply.thenApply` and `thenCatch`.
* Crash in Drafty (caught on Android with Crashlytics): `fmt.at` or `fmt.at + fmt.len` could be out of bounds. It was not a problem with the original JS code. That got copied without the check to Android then replicated to iOS.
* Saving user's email/phone number to DB, otherwise crash on disconnected device.
* Adding/deleting/confirming credentials work now.
* Moved eye icon in password field to the left to make space for ( i ) icon when the field is required but not filled.

### DIFF
--- a/TinodeSDK/MeTopic.swift
+++ b/TinodeSDK/MeTopic.swift
@@ -47,7 +47,7 @@ open class MeTopic<DP: Codable & Mergeable>: Topic<DP, PrivateType, DP, PrivateT
         }
 
         let cred = Credential(meth: meth, val: val)
-        return try! tnd.delCredential(cred: cred).then(
+        return tnd.delCredential(cred: cred).then(
             onSuccess: { [weak self] msg in
                 let idx = self?.find(cred: cred, anyUnconfirmed: false) ?? -1
                 if idx >= 0 {
@@ -57,8 +57,7 @@ open class MeTopic<DP: Codable & Mergeable>: Topic<DP, PrivateType, DP, PrivateT
                 // Notify listeners
                 (self?.listener as! MeListener).onCredUpdated(cred: self?.creds)
                 return nil
-            },
-            onFailure: nil)
+            })
     }
 
     private func find(cred other: Credential, anyUnconfirmed: Bool) -> Int {
--- a/TinodeSDK/MeTopic.swift
+++ b/TinodeSDK/MeTopic.swift
@@ -9,13 +9,16 @@ import Foundation
 
 
 open class MeTopic<DP: Codable & Mergeable>: Topic<DP, PrivateType, DP, PrivateType> {
-    open class MeListener: Topic<DP, PrivateType, DP, PrivateType>.Listener {
+    open class Listener: Topic<DP, PrivateType, DP, PrivateType>.Listener {
         // Called when user credentials are updated.
         open func onCredUpdated(cred: [Credential]?) {}
     }
 
     private var credentials: [Credential]? = nil
 
+    public init(tinode: Tinode?) {
+        super.init(tinode: tinode, name: Tinode.kTopicMe, l: nil)
+    }
     public init(tinode: Tinode?, l: MeTopic<DP>.Listener?) {
         super.init(tinode: tinode, name: Tinode.kTopicMe, l: l)
     }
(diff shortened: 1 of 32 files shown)


TinodeSDK/MeTopic.swift
```diff
@@ -7,15 +7,43 @@
...
     public init(tinode: Tinode?, desc: Description<DP, PrivateType>) {
         super.init(tinode: tinode, name: Tinode.kTopicMe, desc: desc)
     }
 
+    public func serializeCreds() -> String? {
+        guard let c = self.creds else { return nil }
+        return Tinode.serializeObject(c)
+    }
+    public func deserializeCreds(from data: String?) -> Bool {
+        if let c: [Credential]? = Tinode.deserializeObject(from: data) {
```
aforge, line 40: Why `c: [Credential]?` ?
Can you just `if let c = Tinode.deserializeObject(from: data) {`?

TinodeSDK/MeTopic.swift
```diff
@@ -25,6 +53,55 @@ open class MeTopic<DP: Codable & Mergeable>: Topic<DP, PrivateType, DP, PrivateT
...
+    public func delCredential(meth: String, val: String) -> PromisedReply<ServerMessage> {
+        return delCredential(Credential(meth: meth, val: val))
+    }
+
+    public func delCredential(_ cred: Credential) -> PromisedReply<ServerMessage> {
+        let tnd = tinode!
+
+        return tnd.delCredential(cred: cred)
+            .thenApply { [weak self] msg in
+                let idx = self?.find(cred: cred, anyUnconfirmed: false) ?? -1
```
aforge, line 76: Consider `guard let me = self else { return nil }`.

TinodeSDK/MeTopic.swift
```diff
@@ -25,6 +53,55 @@ open class MeTopic<DP: Codable & Mergeable>: Topic<DP, PrivateType, DP, PrivateT
...
+                return nil
+            }
+    }
+
+    public func confirmCred(meth: String, response: String) ->PromisedReply<ServerMessage> {
+        let cred = Credential(meth: meth, val: nil, resp: response, params: nil)
+        return setMeta(meta: MsgSetMeta(desc: nil, sub: nil, tags: nil, cred: cred));
+    }
+
+    private func find(cred other: Credential, anyUnconfirmed: Bool) -> Int {
```
aforge, line 93: Rename credIndex maybe?

TinodeSDK/MeTopic.swift
```diff
@@ -25,6 +53,55 @@ open class MeTopic<DP: Codable & Mergeable>: Topic<DP, PrivateType, DP, PrivateT
...
+        guard let creds = creds else { return -1 }
+        var i = 0
+        for c in creds {
+            if c.meth == other.meth && ((anyUnconfirmed && !c.isDone) || c.val == other.val) {
```
aforge, line 97: You can change the explicit loop to
`return creds.firstIndex { $0.meth == other.meth && ((anyUnconfirmed && !$0.isDone) || $0.val == other.val) }`

TinodeSDK/MeTopic.swift
```diff
@@ -193,4 +286,63 @@ open class MeTopic<DP: Codable & Mergeable>: Topic<DP, PrivateType, DP, PrivateT
...
+                        idx = find(cred: cred, anyUnconfirmed: true)
+                        if idx >= 0 {
+                            // Remove previous unconfirmed credential.
+                            credentials!.remove(at: idx)
+                        }
+                    }
+                    credentials!.append(cred)
+                } else {
+                    // Found. Maybe change 'done' status.
+                    credentials?[idx].done = true
```
aforge, line 313: Should this be `credentials?[idx].done = cred.done`?

TinodeSDK/MeTopic.swift
or-else: It produces `Generic parameter 'T' could not be inferred` without the `: [Credential]?`

TinodeSDK/MeTopic.swift
or-else: done

TinodeSDK/MeTopic.swift
or-else: Let's use `findCredIndex`. A similar name is also used in Android and web. I would prefer to have `find` in it for consistency.

TinodeSDK/MeTopic.swift
or-else: Done. Thanks!

TinodeSDK/MeTopic.swift
or-else: Yes, thanks!

PR #147, opened 2020-05-04
Suppress silent push notifications when the app is in the foreground.

aforge:


or-else:
There is also this:
https://github.com/tinode/ios/blob/master/TinodiosNSExtension/NotificationService.swift#L46

Should it be removed?

aforge:
> There is also this:
> https://github.com/tinode/ios/blob/master/TinodiosNSExtension/NotificationService.swift#L46
> 
> Should it be removed?

Yes. Thanks.

PR #148, opened 2020-05-12
Split AccountSettingVC into 5 screens

or-else:
AccountSetting VC was refactored into five screens to support adding more options (specifically Blocked accounts need to be added to keep iOS on par with web and Android).

A couple of bugs were fixed: incorrect handling of future and making UI changes from background thread.


Tinodios/SettingsGeneralViewController.swift
```diff
@@ -0,0 +1,12 @@
+//
+//  SettingsGeneralViewController.swift
+//  Tinodios
+//
+//  Copyright © 2020 Tinode. All rights reserved.
+//
+
+import UIKit
+
+class SettingsPersonalViewController: UITableViewController {
```
aforge, line 10: Why do you need to declare it here?
It's also declared in Tinodios/SettingsPersonalViewController.swift.

or-else: Fixed. It was some kind of fluke in Xcode - the file was not showing in the UI.

Tinodios/SettingsPersonalViewController.swift
```diff
@@ -0,0 +1,294 @@
...
+            actionTarget: self)
+        UiUtils.setupTapRecognizer(
+            forView: manageContacts,
+            action: #selector(SettingsPersonalViewController.addContactClicked),
+            actionTarget: self)
+
+        self.imagePicker = ImagePicker(presentationController: self, delegate: self, editable: true)
+    }
+
+    private func reloadData() {
```
aforge, line 61: Can you also refresh the contacts table please?

Tinodios/SettingsSecurityViewController.swift
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    private func logout() {
+        Cache.log.info("SettingsSecurityVC - logging out")
+        UiUtils.logoutAndRouteToLoginVC()
+    }
+
+    private func deleteAccount() {
+        Cache.log.info("SettingsSecurityVC - deleting account")
+        tinode.delCurrentUser(hard: true).thenApply { _ in
```
aforge, line 202: Can you add a failure handler please?

Tinodios/SettingsPersonalViewController.swift
or-else: There are some problems with contacts which I want to address in the next PR. For instance, contacts don't show up at all without a connection.

Tinodios/SettingsSecurityViewController.swift
or-else: done

PR #150, opened 2020-05-14
Handling of blocked chats

or-else:
* Blocked chats link is shown on Account and Security page.
* List of blocked chats with an option to unblock and delete.
* Some bug fixes.

### DIFF
--- a/Tinodios/ChatListInteractor.swift
+++ b/Tinodios/ChatListInteractor.swift
@@ -157,9 +157,9 @@ class ChatListInteractor: ChatListBusinessLogic, ChatListDataStore {
                 self?.loadAndPresentTopics()
                 return nil
             },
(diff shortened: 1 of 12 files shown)


Tinodios/ChatListInteractor.swift
```diff
@@ -157,9 +157,9 @@ class ChatListInteractor: ChatListBusinessLogic, ChatListDataStore {
                 self?.loadAndPresentTopics()
                 return nil
             },
-            onFailure: { _ in
+            onFailure: { err in
```
aforge, line 160: You can perhaps change it to `UiUtils.ToastFailureHandler`.
The `Cache.log.error` statement is redundant.

or-else: done

PR #155, opened 2020-06-07
Apply typing notifications and read receipts settings.

aforge:
The account notifications settings are currently unused.
Change it.

### DIFF
--- a/Tinodios/MessageInteractor.swift
+++ b/Tinodios/MessageInteractor.swift
@@ -11,7 +11,7 @@ import TinodiosDB
 
 protocol MessageBusinessLogic: class {
     @discardableResult
-    func setup(topicName: String?) -> Bool
+    func setup(topicName: String?, sendReadReceipts: Bool) -> Bool
     @discardableResult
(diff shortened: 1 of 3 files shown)


Tinodios/MessageInteractor.swift
```diff
@@ -199,6 +201,7 @@ class MessageInteractor: DefaultComTopic.Listener, MessageBusinessLogic, Message
         )
     }
     func sendReadNotification(explicitSeq: Int? = nil, when deadline: DispatchTime) {
+        guard self.sendReadReceipts else { return }
```
or-else, line 204: Why do it differently than `sendTypingNotifications`? You check for `sendTypingNotifications` in `MessageViewController+...`, but `sendReadReceipts` is checked in `MessageInteractor`.

aforge: That's because `sendReadReceipts ` is called from two places: `MessageViewController.viewDidAppear` and `MessageInteractor.onData`. So I added this check only in `MessageInteractor.sendReadNotification`.

PR #162, opened 2020-08-31
Support for channels

or-else:


Tinodios/MessageViewController.swift
```diff
@@ -1108,9 +1108,12 @@ extension MessageViewController : MessageCellDelegate {
         }
 
         // Set up the shared UIMenuController
-        let copyMenuItem = MessageMenuItem(title: NSLocalizedString("Copy", comment: "Menu item"), action: #selector(copyMessageContent(sender:)), seqId: cell.seqId)
-        let deleteMenuItem = MessageMenuItem(title: NSLocalizedString("Delete", comment: "Menu item"), action: #selector(deleteMessage(sender:)), seqId: cell.seqId)
-        UIMenuController.shared.menuItems = [copyMenuItem, deleteMenuItem]
+        var menuItems: [MessageMenuItem] = []
+        menuItems.append(MessageMenuItem(title: NSLocalizedString("Copy", comment: "Menu item"), action: #selector(copyMessageContent(sender:)), seqId: cell.seqId))
+        if !(topic?.isChannel ?? false) {
```
aforge, line 1113: I would change `false` to `true` here.
This clause may be hit only if `topic` is nil - in which case we want to make available as few options as possible.

Tinodios/MessageViewController.swift
```diff
@@ -1108,9 +1108,12 @@ extension MessageViewController : MessageCellDelegate {
...
+            // Channel users cannot delete messages.
+            menuItems.append(MessageMenuItem(title: NSLocalizedString("Delete", comment: "Menu item"), action: #selector(deleteMessage(sender:)), seqId: cell.seqId))        }
```
aforge, line 1115: Move } to the next line for better readability?

or-else:
> In the Find tabset, it's probably a good idea to rename the "New Group" tab "New Channel"

That tab is used to create both groups and channels. Channel is a type of a group. I don't think it's a good idea to rename the tab. The default `This is a channel [on]` is wrong though. I'll make it `[off]`. 

> The following scenario leads to a server error:

Good catch. This is purely server-side: first a logical problem with default permissions for members, second logical problem with search results for existing members.

Tinodios/MessageViewController.swift
or-else: done

Tinodios/MessageViewController.swift
or-else: done

PR #163, opened 2020-09-17
Topic redirects

or-else:
Also removed the online (green or gray) dot for channels.

### DIFF
--- a/Tinodios/MessageInteractor.swift
+++ b/Tinodios/MessageInteractor.swift
@@ -135,7 +135,7 @@ class MessageInteractor: DefaultComTopic.Listener, MessageBusinessLogic, Message
         topic.subscribe(set: nil, get: builder.build()).then(
                 onSuccess: { [weak self] msg in
                     // Check for topic redirects.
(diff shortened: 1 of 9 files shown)


Tinodios/MessageInteractor.swift
```diff
@@ -133,7 +133,17 @@ class MessageInteractor: DefaultComTopic.Listener, MessageBusinessLogic, Message
             builder = builder.withTags()
         }
         topic.subscribe(set: nil, get: builder.build()).then(
-                onSuccess: { [weak self] _ in
+                onSuccess: { [weak self] msg in
+                    // Check for topic redirects.
+                    if let ctrl = msg?.ctrl, ctrl.code == 303 {
```
aforge, line 138: Consider introducing a constant for 303 in TinodeSDK.ServerMessage.

aforge (approved):

Tinodios/MessageInteractor.swift
or-else: done

PR #164, opened 2020-11-23
Download and display of out-of-band images

or-else:
Download and display of images works, uploading is not yet implemented.
Using Kingfisher because it not just downloads but also manages the cache.

Also updated pods because Crashlytics stopped working and required an update.
Updated minimum iOS version to 12 because something (don't remember what exactly) did not work below 12. 12 covers 98% of devices.


Tinodios/AttributedStringFormatter.swift
```diff
@@ -417,15 +433,51 @@ class AttributedStringFormatter: DraftyFormatter {
...
+                let url: URL?
+                if let ref = attachment.ref {
+                    url = URL(string: ref, relativeTo: tinode.baseURL(useWebsocketProtocol: false))
+                } else {
+                    url = nil
+                }
+                let wrapper = url == nil ? NSTextAttachment() : AsyncTextAttachment(url: url!)
+
+                var image: UIImage?
+                if let bits = attachment.bits, let preveiw = UIImage(data: bits) {
```
aforge, line 447: s/preveiw/preview

aforge (commented):
How'd you test it?

Tinodios/AttributedStringFormatter.swift
or-else: Done

or-else:
> How'd you test it?

Web and Android can upload images out of band already.

Tinodios/widgets/AsyncTextAttachment.swift
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    public func startDownload(onError errorImage: UIImage) {
+        let modifier = AnyModifier { request in
+            var request = request
+            LargeFileHelper.addCommonHeaders(to: &request, using: Cache.getTinode())
+            return request
+        }
+
+        KingfisherManager.shared.retrieveImage(with: url.downloadURL, options: [.requestModifier(modifier)], completionHandler: { result in
```
aforge, line 39: This caches the image, correct?
When I ran this code under debugger, it paused at this line multiple times when rendering a message.

aforge (approved):

Tinodios/widgets/AsyncTextAttachment.swift
or-else: It looks like it caches the images: there is just one request to the server. The caching is also listed as a feature: https://onevcat.github.io/Kingfisher/index.html

or-else: The only reason it's used is because of the cache. Otherwise it's pretty simple to implement async downloads.

aforge: Cool. Let's move forward with it then.

PR #165, opened 2020-12-06
Image uploads

or-else:
I've implemented the same functionality as in web and Android for uploading large image and sending the URL.
I've refactored the progress indicator: it's now a bar instead of a circle. It works OK for both attachments and images. The circle looked odd with images. 

There is at least one unfixed bug (unrelated to uploads): when message sending fails and I delete the failed unsent message, the app tries to send {del} request to the server, fails, and that {del}-fail cycle is stuck forever.


Tinodios/AttributedStringFormatter.swift
```diff
@@ -387,21 +389,35 @@ class AttributedStringFormatter: DraftyFormatter {
...
+
+                second.endEditing()
+                attributed.append(second)
             }
 
             if isValid {
                 // Insert linebreak then a clickable [↓ save] line
                 attributed.append(NSAttributedString(string: "\u{2009}\n", attributes: [NSAttributedString.Key.font : baseFont]))
 
-                // \u{2009} because iOS is buggy and bugs go unfixed for years.
```
aforge, line 397: Why remove this comment?
The context may be helpful in the future.

Tinodios/LargeFileHelper.swift
```diff
@@ -1,64 +1,94 @@
...
+    public func finished(msg: ServerMessage?, err: Error?) {
+        self.finalCb?(msg, err)
+        self.finalCb = nil
+    }
 }
 
-class LargeFileHelper: NSObject {
-    static let kBoundary = "*****\(Date().millisecondsSince1970)*****"
+public class LargeFileHelper: NSObject {
+    static let kBoundary = "*****\(Int64((Date().timeIntervalSince1970 * 1000.0).rounded()))*****"
```
aforge, line 60: `millisecondsSince1970` is defined as `(timeIntervalSince1970 * 1000.0).rounded()`.

Tinodios/LargeFileHelper.swift
```diff
@@ -1,64 +1,94 @@
...
+    convenience init(with tinode: Tinode) {
         let config = URLSessionConfiguration.background(withIdentifier: Bundle.main.bundleIdentifier!)
-        self.init(config: config)
+        self.init(with: tinode, config: config)
     }
+
     public static func addCommonHeaders(to request: inout URLRequest, using tinode: Tinode) {
         request.addValue(tinode.apiKey, forHTTPHeaderField: "X-Tinode-APIKey")
-        request.addValue("Token \(tinode.authToken!)", forHTTPHeaderField: "X-Tinode-Auth")
+        if tinode.isConnectionAuthenticated {
```
aforge, line 80: Won't the upload fail if the connection isn't authenticated?
Maybe return a `bool` from this function and handle it in the caller?

Tinodios/AttributedStringFormatter.swift
or-else: I moved it up to the top of the code block.

Tinodios/LargeFileHelper.swift
or-else: I tried to move LargeFileHelper to the SDK, failed, moved it back and this was lost in the process. Sorry. I put it back.

Tinodios/LargeFileHelper.swift
or-else: I think it's not appropriate to return anything from this call. The check `isConnectionAuthenticated` is available to the caller.

or-else:
It looks like the bug with deletion of failed messages is also present in Android.

aforge:
> It looks like the bug with deletion of failed messages is also present in Android. I'm going to try to fix them both before release.

Sounds good to me.

aforge (approved):

PR #166, opened 2020-12-08
Fix problem with deletion of failed messages, some refactoring

or-else:


TinodiosDB/BaseDb.swift
```diff
@@ -11,27 +11,36 @@ import TinodeSDK
...
-    // Object is received by the server.
-    public static let kStatusSynced = 4
+    public static let kSchemaVersion: Int32 = 106
+
+    // Object statuses. Values are incremented by 10 to make it easier to add new statuses.
+    public enum Status: Int, Comparable {
+        // Status undefined/not set.
+        case undefined = 0
+        // Object is not ready to be sent to the server.
+        case draft = 10
```
aforge, line 21: I'd perhaps keep the values the same for backward compatibility reasons
(the phones that have the app installed persisted the old enum values in their local sqlitedbs).

aforge (approved):

TinodiosDB/BaseDb.swift
or-else: I can't keep them the same: `.failed` has to be less then `.sent`. I've changed the DB version, everything will be re-downloaded.

PR #168, opened 2021-06-17
Make archived chats button display in the header.

aforge:
The footer doesn't stick at the screen bottom but rather scrolls
down with the table content. It makes the archived chats button
invisible and hardly noticeable when the user has many subscriptions.

### DIFF
--- a/Tinodios/ChatListViewController.swift
+++ b/Tinodios/ChatListViewController.swift
@@ -29,8 +29,8 @@ class ChatListViewController: UITableViewController, ChatListDisplayLogic {
     // Index of contacts: name => position in topics
     var rowIndex: [String : Int] = [:]
     var router: ChatListRoutingLogic?
-    // Archived chats footer
-    var archivedChatsFooter: UIView?
+    // Archived chats header.
+    var archivedChatsHeader: UIView?
 
     private func setup() {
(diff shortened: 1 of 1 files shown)


aforge:
Before 
Simulator Screen Shot - iPhone 8 - 2021-06-17 at 13 14 16
After 
Simulator Screen Shot - iPhone 8 - 2021-06-17 at 13 11 58

or-else:
Let's not do it. The current position as the last item in the list follows the UX guidelines: it's a rarely used button placed in the least valuable position. The importance of the feature is supported by the UI position: low importance -> not easily accessible position. You don't want archived chats link to take the most valuable position on the screen because you don't want to see these chats, that's why they are archived. There is no need to make them immediately accessible. Take a look at WhatsApp UI.

aforge:
> Let's not do it. The current position as the last item in the list follows the UX principles: it's a rarely used button placed in the least valuable position. The importance of the feature is supported by the UI position: low importance -> not easily accessible position. You don't want archived chats link to take the most valuable position on the screen because you don't want to see these chats, that's why they are archived. There is no need to make them immediately accessible. Take a look at WhatsApp UI.

Sounds good. Am going to drop this PR.

PR #176, opened 2021-10-22
Basic changes for rendering QQ Drafty elements in Tinodios.

aforge:
This PR deals with the rendering part only since it's already too big. Replies are WIP.

1. Moved `DraftyFormatter`'s associated type `Node` out of the enclosing protocol (protocols with an associated type are too hard to work with when using generics).
2. Refactored the logic which turns `Drafty` into a tree of `Drafty.Span`s and traverses it into a separate `Drafty.SpanTreeProcessor` class. Now `Drafty.format()` and `Drafty.preview()` may implement their own processors and apply their own logic in them by subclassing `Drafty.SpanTreeProcessor`.
3. Quotes (`QQ`s) are rendered via `QuotedAttachment`s - these are similar to the `DraftyButtonAttachment`. The quoted `Drafty` object is rendered in an image object which subsequently is embedded in a `NSTextAttachment` in the parent `Drafty` object (unfortunately, Apple's NSAttributedString does not have anything similar to the Android's Spans).

### DIFF
--- a/TinodeSDK/model/Drafty.swift
+++ b/TinodeSDK/model/Drafty.swift
@@ -1054,6 +1054,8 @@ open class Drafty: Codable, CustomStringConvertible, Equatable {
         let proc = PreviewProcessor<T>()
         let tree = proc.process(contentOf: self, using: tranformer) as! Span
         tree.clip(upToMax: length)
+        // Remove leading whitespaces.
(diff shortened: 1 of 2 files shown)


or-else:
It would be nice to port tests too.

PR #179, opened 2022-06-06
Video calls iOS

aforge:
Basic functionality should be working


or-else:
Looks OK at the first glance. Let's merge and then fix.

I made two material changes in Android compare to your code: 
1. Instead of trying to read the latest version of each message from the main table, I mutated the main message with every change. The versions were also saved to the main table but with a reference ID, which made it easy to filter them out. The versions of the overwritten message content were stored into a separate table `EditHistory` in case we want ot show edit history. It simplified the code (no need for in-memory cursor) and fixed a couple of bugs (like replying to a wrong message version and incorrect message bubble).
2. I removed VC-related custom variables from formatter classes and moved all the data they contained to VC entity itself. That required rewriting the VC drafty content with the values from the header. I added such rewrites to two places: where the message is received and where it's sent.

It might be a good idea to do it the same way here too.

BTW, the cache management in iOS is busted: it does not work when trying to fetch older messages. I'm trying to fix it now. 
It's probably busted in Android too.

or-else (approved):

aforge:
> Looks OK at the first glance. Let's merge and then fix.
> 
> I made two material changes in Android compare to your code:
> 
> 1. Instead of trying to read the latest version of each message from the main table, I mutated the main message with every change. The versions were also saved to the main table but with a reference ID, which made it easy to filter them out. The versions of the overwritten message content were stored into a separate table `EditHistory` in case we want ot show edit history. It simplified the code (no need for in-memory cursor) and fixed a couple of bugs (like replying to a wrong message version and incorrect message bubble).
> 2. I removed VC-related custom variables from formatter classes and moved all the data they contained to VC entity itself. That required rewriting the VC drafty content with the values from the header. I added such rewrites to two places: where the message is received and where it's sent.
> 
> It might be a good idea to do it the same way here too.
> 
> BTW, the cache management in iOS is busted: it does not work when trying to fetch older messages. I'm trying to fix it now. It's probably busted in Android too.

Sounds good. I'll follow up with the similar changes in iOS.
Merging now.

PR #180, opened 2022-06-14
Remove EditHistoryDb

aforge:
This PR changes the logic of message replacement as follows:
- EditHistoryDb is gone.
- message `content` and `head` are never updated
- added a new column `replaced_by_seq` to MessageDb which represents `seq` of the message that replaces the current row. This allows for easy latest version lookups.


or-else (approved):
LGTM.

A consideration though. We need to solve two problems with this:

1. Select the latest version of the message keeping messages in the right order sorted by `seq`.
2. Keep edit history.

This implementation is "light insert - heavy select". I was thinking about doing it in the opposite way "heavy insert - light select" because selects are frequent, inserts are rare.

I was thinking about mutating the `seq` field making it "effective seq" instead of the field matching the server and have an `originalSeq`. Or maybe keep `seq` as is and create a field `effectiveSeq`. 

When a new message A is received (assuming `seq` is kept as is + new field `effectiveSeq` is used for filtering and sorting):
1. Try to locate an existing message B with `effectiveSeq` equal to the `seq` of the new message A.
1.1 If such message B is found, compare `B.seq` with the `A.seq`.
1.1.1 `B.seq` > `A.seq` means message A has already been replaced by B: insert A with  `A.replacedWith=B.seq`, `A.effectiveSeq = NULL`. 
1.1.2 `B.seq` < `A.seq` means A is a replacement for B: update B with `B.replacedWith=A.seq` and `B.effectiveSeq=NULL`; insert A with `A.effectiveSeq = A.seq`.
1.2 B is not found: just set `A.effectiveSeq = A.seq` and insert A.

In this scheme select can be a plain `SELECT * WHERE effectiveSeq IS NOT NULL ORDER BY effectiveSeq`, no joins. An INSERT is a bit heavy but not too horrible. 

What do you think?

or-else:
BTW, SQLite has a thing called [Partial Index](https://www.sqlite.org/partialindex.html). It can be used on `effectiveSeq` to make it unique non - null:
`CREATE UNIQUE INDEX ... ON messages(table_id,effectiveSeq)  WHERE effectiveSeq IS NOT NULL`

or-else:
This approach (and I think your approach too) won't work correctly in case of "A replaced by B replaced by C, when only A and C are downloaded". I don't think it's a big deal though. We can consider such chained replacements a client bug.

aforge:
> 

Yeah, I think `effective_seq` is a reasonable solution here. It can be used to avoid incorrectly rewriting `head` and `content` when replacement messages arrive in reverse chronologic order.

Additionally, we can also add `effective_head` and `effective_content` (that come from replacement messages) in order to keep `head` and `content` intact.

Upsides:
- all history is kept within the `messages` table.
- the `select` remains light as does the `insert`.
Downsides:
- a bit of denormalization with `effective_{head,content}` fields being stored twice

LMK what you think.

> This approach (and I think your approach too) won't work correctly in case of "A replaced by B replaced by C, when only A > and C are downloaded". I don't think it's a big deal though. We can consider such chained replacements a client bug.

What is the problem here?
We don't send replacements for replacements. C is supposed to replace A in this scheme so the new content of A should be fine.

If we didn't download B, it just won't show up in A's edit history.

or-else:
> Additionally, we can also add effective_head and effective_content (that come from replacement messages) in order to keep head and content intact

I don't understand this. With `effective_seq` you don't need to rewrite anything but `effective_seq` and `replaced_with`. I think there is still some misunderstanding. Should we voice chat? 

> We don't send replacements for replacements.

Right. But someone might create a client which does it. We can just treat such client as buggy.

aforge:
> > Additionally, we can also add effective_head and effective_content (that come from replacement messages) in order to keep head and content intact
> 
> I don't understand this. With `effective_seq` you don't need to rewrite anything but `effective_seq` and `replaced_with`. I think there is still some misunderstanding. Should we voice chat?
> 
> > We don't send replacements for replacements.
> 
> Right. But someone might create a client which does it. We can just treat such client as buggy.

Changed to `effective_seq`. PTAL

TinodiosDB/MessageDb.swift
```diff
@@ -193,17 +156,26 @@ public class MessageDb {
         }
         do {
             try db.savepoint("MessageDb.insert") {
-                if try insertRaw(topic: topic, msg: msg, withSeqId: nil) > 0,  // use seq id from msg
-                   let replacedSeq = msg.replacesSeq, replacedSeq > 0 {
-                    try replaceMessage(onTopic: topic, withOrigSeqId: replacedSeq, withMsg: msg)
+                var effSeq: Int? = msg.replacesSeq ?? msg.seqId
```
or-else, line 159: Why optional `Int?`? `msg.seqId` is never `nil`, consequently `effSeq` is never `nil`.

or-else (approved):

TinodiosDB/MessageDb.swift
or-else: Oh, I see why. Resolved.

or-else:
BTW, I think a non-unique index on `topicId + effectiveSeq` would be useful. 
It would be better to create a partial unique index on it, but SQLite.swift does not support partial indexes. It can be created with plain SQL though.

TinodiosDB/MessageDb.swift
```diff
@@ -386,8 +376,8 @@ public class MessageDb {
 
         var query = self.table
         query = forward ?
-            query.filter(self.topicId == topicId && self.seq > from && self.replSeq == nil) :
-            query.filter(self.topicId == topicId && self.seq < from && self.replSeq == nil)
+            query.filter(self.topicId == topicId && self.seq > from && self.effectiveSeq != nil) :
```
or-else, line 379: I think it should be 
```swift
query.filter(self.topicId == topicId && self.effectiveSeq > from)
```

TinodiosDB/MessageDb.swift
```diff
@@ -386,8 +376,8 @@ public class MessageDb {
...
+            query.filter(self.topicId == topicId && self.seq < from && self.effectiveSeq != nil)
```
or-else, line 380: Likewise, 
```swift
query.filter(self.topicId == topicId && self.effectiveSeq < from)
```

TinodiosDB/MessageDb.swift
```diff
@@ -386,8 +376,8 @@ public class MessageDb {
...
         query = (forward ? query.order(self.seq.asc) : query.order(self.seq.desc)).limit(limit)
```
or-else, line 381: `query.order(self.effectiveSeq.xxx)`

aforge: Done.

TinodiosDB/MessageDb.swift
aforge: Done.

TinodiosDB/MessageDb.swift
aforge: Done.

aforge:
> BTW, I think a non-unique index on `topicId + effectiveSeq` would be useful. It would be better to create a partial unique index on it, but SQLite.swift does not support partial indexes. It can be created with plain SQL though.

Added.

TinodiosDB/MessageDb.swift
```diff
@@ -486,17 +481,7 @@ public class MessageDb {
...
-            self.table[self.userId],
-            self.table[self.status],
-            self.table[self.sender],
-            self.table[self.ts],
-            self.table[self.seq],
-            self.table[self.high],
-            self.table[self.delId],
-            self.table[self.head],
-            self.table[self.content],
+            self.table[*],
```
or-else, line 484: This too needs to be updated. The current `queryLatest` selects latest version of the message instead of latest message.

`self.table[self.seq] < messages2[self.seq]` -> `self.table[self.effectiveSeq] < messages2[self.effectiveSeq]` and `self.effectiveSeq != nil` has to be somewhere.

PR #181, opened 2022-07-25
Add handling of .voip pushes

or-else:


Tinodios/AppDelegate.swift
```diff
@@ -255,3 +243,50 @@ extension AppDelegate: MessagingDelegate {
...
+    func pushRegistry(_ registry: PKPushRegistry, didInvalidatePushTokenFor type: PKPushType) {
+        print("PK must invalidate token")
+    }
+
+    // VoIP push notification recived.
+    func pushRegistry(_ registry: PKPushRegistry, didReceiveIncomingPushWith payload: PKPushPayload, for type: PKPushType, completion: @escaping () -> Void) {
+        print("PK push %s", payload.debugDescription)
+
+        if type != .voIP {
+            return
```
aforge, line 261: Call completion()?

aforge (approved):

Tinodios/AppDelegate.swift
or-else: done

PR #184, opened 2022-11-04
<fix> blurred avatar while in personalsetting scene

yuyii:
I found my avatar is blurred if I use custom image for my avatar. And I found the problem is in RoundImageView.swift.

The image would be overrided by base64 data while icon set. I just added the function param: isAvatar . 
And it can differ from list icon and avatar.

It's useful for me.

### DIFF
--- a/Tinodios/AccountSettingsViewController.swift
+++ b/Tinodios/AccountSettingsViewController.swift
@@ -42,7 +42,7 @@ class AccountSettingsViewController: UITableViewController {
         self.userNameLabel.text = me.pub?.fn ?? NSLocalizedString("Unknown", comment: "Placeholder for missing user name")
 
         // Avatar.
-        self.avatarImageView.set(pub: me.pub, id: self.tinode.myUid, deleted: false)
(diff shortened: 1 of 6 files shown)


or-else:
Good catch, thanks. But the fix seems excessive. The `RoundImageView` is used to display avatars only and nothing but avatars. You don't need to add `isAvatar: Bool = false` to `RoundImageView.set()`. It's safe to assume that `isAvatar` is always `true`.

or-else:
The logic here is that the thumbnail is assigned first and displayed until the full image is downloaded. But it looks like in your case `Kingfisher` downloads the image before direct assignment of a thumbnail. It's a data race. 

I.e. this line executes first
https://github.com/tinode/ios/blob/8c100634f8bdc347b0dc92cc4aa796517041de0d/Tinodios/widgets/RoundImageView.swift#L111
Then this is executed
https://github.com/tinode/ios/blob/8c100634f8bdc347b0dc92cc4aa796517041de0d/Tinodios/widgets/RoundImageView.swift#L125

when it should be the other way.

or-else:
How about this instead? 

https://github.com/tinode/ios/pull/185

yuyii:
I see. That's good.

PR #185, opened 2022-11-04
fix for data race in displaying an avatar

or-else:


Tinodios/widgets/RoundImageView.swift
```diff
@@ -100,30 +100,15 @@ public class RoundImageView: UIImageView {
...
-                    self.initials = nil
-                    self.backgroundColor = nil
-                    self.image = deleted ? value.image.noir : value.image
-                }
-                // Ignoring the error: just keep the placeholder image.
-            })
-        }
-
         if let icon = pub?.photo?.image {
+            // Use thumbnail, if preset.
```
aforge, line 104: s/preset/present

or-else: done

PR #190, opened 2023-01-04
Editing: make sure message can be edited only by author.

aforge:


or-else:
This works correctly if the original message is already in the DB when the replacement is received, but what if replacement is received first?

aforge:
> This works correctly if the original message is already in the DB when the replacement is received, but what if replacement is received first?

Thanks for pointing out. Fixed.

or-else (approved):

PR #192, opened 2023-01-17
Audio only calls.

aforge:
Should be generally working.
Let me know if you find issues.

### DIFF
--- a/TinodeSDK/model/ServerMessages.swift
+++ b/TinodeSDK/model/ServerMessages.swift
@@ -188,6 +188,11 @@ open class MsgServerData: Codable {
         return WebRTC(rawValue: str)
     }
 
+    /// Returns true if the message is marked "audio-only" (makes sense only if it's a call).
+    public var isAudioOnlyCall: Bool {
+        return self.head?["aonly"]?.asBool() ?? false
+    }
+
     // Testing only.
     internal init(id: String?, topic: String?, from: String?, ts: Date?,
                   head: [String: JSONValue]?, seq: Int?, content: Drafty?) {
--- a/Tinodios/AppDelegate.swift
+++ b/Tinodios/AppDelegate.swift
@@ -54,13 +54,14 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             guard let msg = topic.getMessage(byEffectiveSeq: seqId) as? StoredMessage,
                   msg.webrtcCallState == callState else { return }
 
+            let isAudioOnly = data.isAudioOnlyCall
             Cache.log.info("Call (topic: %@, seq: %d): processing event %@", topicName, seqId, String(reflecting: callState))
             switch callState {
             case .kStarted:
                 // It is a legit incoming call. Start it.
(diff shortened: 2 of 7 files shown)


or-else:
iOS works well, no technical issues. But there is a problem on the web: the video button does not work when calling with iOS. It works with android.

This UI looks crowded:

Screenshot 2023-01-17 at 10 23 36

Maybe show just one button, either video or audio and save the type of call (with video or without) on the client and show the button for the same call as before.

The video icon in the toolbar is likely to be a problem during app review. Apple states that the icon can only be used for FaceTime. I'll create a different icon. 

Screenshot 2023-01-17 at 10 35 08

or-else:
video-camera-outline

or-else:
Does not work with the web. Web gets a bunch of these errors:
```
TCPeerConnectionIceErrorEvent {isTrusted: true, address: '192.168.5.x', port: 57820, hostCandidate: '192.168.5.x:57820', url: 'turn:openrelay.metered.ca:443?transport=tcp', …}
isTrusted: trueaddress: "192.168.5.x"
bubbles: false
cancelBubble: false
cancelable: false
composed: false
currentTarget: RTCPeerConnection {
  onremovetrack: null, 
  localDescription: RTCSessionDescription, 
  currentLocalDescription: RTCSessionDescription, 
  pendingLocalDescription: null, remoteDescription: RTCSessionDescription, …
}
defaultPrevented: false
errorCode: 600
errorText: "Address not associated with the desired network interface."
eventPhase: 0
hostCandidate: "192.168.5.x:57820"
port: 57820
returnValue: true
srcElement: RTCPeerConnection {
  onremovetrack: null, 
  localDescription: RTCSessionDescription, 
  currentLocalDescription: RTCSessionDescription, 
  pendingLocalDescription: null, 
  remoteDescription: RTCSessionDescription, …
}
target: RTCPeerConnection {
  onremovetrack: null, 
  localDescription: RTCSessionDescription, 
  currentLocalDescription: RTCSessionDescription, 
  pendingLocalDescription: null, 
  remoteDescription: RTCSessionDescription, …
}
timeStamp: 32079102.099999905
type: "icecandidateerror"
url: "turn:openrelay.metered.ca:443?transport=tcp"
```

or-else:
Video calls also get one error like that but works (audio and video are fine). Audio-only does not work

or-else:
Minor UI issue: when calling to iOS, the notification on iPhone always says "Tinode Video" even when it's audio-only.

or-else:
The call notification does not always appear. The data message arrives, but no call notification is shown.

or-else:
When calling to or from iOS the mute and video buttons are always disabled, for audio and video+audio
Screenshot 2023-01-18 at 13 13 20
 calls.

or-else:
Xcode shows a lot of records like this:
```
2023-01-18 13:17:11.922176-0800 Tinodios[531:44615] [co.tinode.tinodios] CallVC - audio route change: other - 4
```

or-else:
It seems like audio in audio-only calls is not coming through: iOS -> Android and iOS -> Web don't work. The other way works fine. Video calls work fine both ways.
No other issues.

or-else:
No longer hanging.

PR #193, opened 2023-01-28
Registration by code

aforge:
Sign up, verification, password & credentials change by code.

### DIFF
--- a/TinodeSDK/model/ClientMessages.swift
+++ b/TinodeSDK/model/ClientMessages.swift
@@ -76,6 +76,8 @@ public class Credential: Codable, Comparable, CustomStringConvertible {
 public class MsgClientAcc<Pu: Codable, Pr: Codable>: Codable {
     var id: String?
     var user: String?
+    var tmpscheme: String?
+    var tmpsecret: String?
     var scheme: String?
     var secret: String?
     var login: Bool?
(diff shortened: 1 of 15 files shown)


TinodeSDK/model/ClientMessages.swift
```diff
@@ -76,6 +76,8 @@ public class Credential: Codable, Comparable, CustomStringConvertible {
 public class MsgClientAcc<Pu: Codable, Pr: Codable>: Codable {
     var id: String?
     var user: String?
+    var tmpscheme: String?
+    var tmpsecret: String?
```
or-else, line 80: Hah - no `Token` in `MsgClientAcc`. It means it never worked before! It's not a comment on your current code, just a general observation.

TinodeSDK/model/TheCard.swift
```diff
@@ -150,18 +150,20 @@ public class TheCard: Codable, Mergeable {
 
     public init() {}
 
-    public init(fn: String?, avatar: Photo?) {
+    public init(fn: String?, avatar: Photo?, note: String? = nil) {
         self.fn = fn
         self.photo = avatar
+        self.note = note
```
or-else, line 156: +1!

or-else (approved):
LGTM other than the UI. I'll work on the UI once I finish the webapp.

PR #194, opened 2023-02-03
UI and UX adjustments in new controllers.

or-else:


Tinodios/CredentialsChangeViewController.swift
```diff
@@ -120,21 +118,6 @@ class CredentialsChangeViewController: UITableViewController {
...
     }
 
-    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
-        guard UIApplication.shared.applicationState == .active else {
-            return
-        }
-        self.setInterfaceColors()
-    }
-
-    private func setInterfaceColors() {
```
aforge, line 130: Is it something we don't need anymore?

Tinodios/CredentialsChangeViewController.swift
```diff
@@ -171,7 +154,11 @@ class CredentialsChangeViewController: UITableViewController {
             .then(onSuccess: { msg in
                 if let ctrl = msg?.ctrl, 200 <= ctrl.code && ctrl.code < 300 {
                     self.confirmationSectionVisible = true
-                    DispatchQueue.main.async { UiUtils.showToast(message: "Confirmation code sent", level: .info) }
+                    DispatchQueue.main.async {
+                        self.newTelField.isEnabled = false
```
aforge, line 158: Nice!
Perhaps disable interaction with the entire view:
`self.view.isUserInteractionEnabled = false`

Tinodios/SettingsPersonalViewController.swift
```diff
@@ -34,6 +34,8 @@ class SettingsPersonalViewController: UITableViewController {
     weak var me: DefaultMeTopic!
     private var imagePicker: ImagePicker!
 
+    private let phoneNumberKit = PhoneNumberKit()
```
aforge, line 37: You can use `UiUtils.phoneNumberKit` to avoid creating another object.

aforge (approved):

Tinodios/CredentialsChangeViewController.swift
or-else: I'm using system colors. They are supposed to change automatically between dark and light themes.

Tinodios/CredentialsChangeViewController.swift
or-else: I want to disable just this field, not with the entire view. Am I missing something?

Tinodios/SettingsPersonalViewController.swift
or-else: Fixed

PR #196, opened 2023-03-08
Branding configuration

aforge:
Branding bits.


TinodiosDB/SharedUtils.swift
```diff
@@ -364,11 +364,39 @@ public class SharedUtils {
         task.resume()
     }
 
+    // Identifies device with Tinode server and fetches branding configuration code.
+    public static func identifyAndConfigureBranding() {
+        let url = URL(string: "https://hosts.tinode.co/whoami")!
+        print("Self-identifying with the server. Endpoint: ", url.absoluteString)
+        let task = URLSession.shared.dataTask(with: URLRequest(url: url)) { data, response, error in
```
or-else, line 371: This would probably have to include some extra info, such as phone and iOS versions. It's OK for now.

or-else (approved):

